PR #30, opened 2022-07-10
MNK Level Unlocks

Capritle:
Added the rest of the MNK level checks for the autorotation. Not sure if UnlockedEnhancedBrotherhood is redundant, as looking at the current code I think the chakra would automatically be managed.


awgil:
Thanks!

PR #189, opened 2023-10-17
Thordan Unreal

TheElec:
Current progress is right before Meteors. 

I'm not sure how to properly add the repeating casts to the Adelphel&Janlenoux part, or if its even necessary beyond the mitigation hint.


awgil:
I'm planning to implement new unreal/extreme asap - was on vacation for the last two weeks - so I'll merge this PR and work from it. Thanks!

PR #207, opened 2023-12-16
Add Leviathan Hard module

Taurenkey:
Whorleater Hard mode module

### DIFF
--- a/BossMod/Modules/RealmReborn/Trial/T09WhorleaterH/T09WhorleaterH.cs
+++ b/BossMod/Modules/RealmReborn/Trial/T09WhorleaterH/T09WhorleaterH.cs
@@ -1,0 +1,112 @@
+﻿using BossMod.Endwalker.Extreme.Ex7Zeromus;
+using FFXIVClientStructs.FFXIV.Client.Game.UI;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BossMod.Modules.RealmReborn.Trial.T09WhorleaterH
+{
+    public enum OID : uint
+    {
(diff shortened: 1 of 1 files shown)


BossMod/Modules/RealmReborn/Trial/T09WhorleaterH/T09WhorleaterH.cs
```diff
@@ -0,0 +1,168 @@
+﻿using BossMod.Endwalker.Extreme.Ex7Zeromus;
+using BossMod.ReplayVisualization;
+using FFXIVClientStructs.FFXIV.Client.Game.UI;
```
awgil, line 3: These all are hopefully not needed? Generally modules should not rely on game state, so that replays can be opened in uidev.

BossMod/Modules/RealmReborn/Trial/T09WhorleaterH/T09WhorleaterH.cs
```diff
@@ -0,0 +1,168 @@
...
+        Boss = 0xA67, // Leviathan's Head
+        Tail = 0xA86, // Tail
+        Spume = 0xA85, // Adds for the converter
+        Sahagin = 0xA84, // Hostile adds
+        DangerousSahagins = 0xA83,
+        Converter = 0x1E922A, // Elemental converter
+        HydroshotZone = 0x1E9230 // Hydroshot AoE puddle
+    };
+
+    public enum AID : uint
```
awgil, line 23: I wonder, did you use enum autogeneration from analysis? While it's not strictly necessary, it auto-adds a bunch of info about abilities as comments, and I find it more convenient than writing out values by hand.

PR #217, opened 2024-01-08
Movement lock alternative

xanunderscore:
this doesn't fix the cast buffering issue mentioned in the comment at the top, but it does work for autorotation where we can block movement in advance of the cast starting, tested it with my WIP sam rotation. this is really just convenient for when WASD aren't the user's movement keys, such as for azerty or dvorak users

### DIFF
--- a/BossMod/Autorotation/InputOverride.cs
+++ b/BossMod/Autorotation/InputOverride.cs
@@ -30,6 +30,9 @@ namespace BossMod
         private delegate ref int GetRefValueDelegate(int vkCode);
         private GetRefValueDelegate _getKeyRef;
 
+        private IntPtr _forceDisableMovementPtr;
+        internal unsafe ref int ForceDisableMovement => ref *(int*)(_forceDisableMovementPtr + 4);
+
         public unsafe InputOverride()
         {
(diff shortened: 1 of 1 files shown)


xanunderscore:
while testing this, I noticed that holding a movement key during movement lock can prevent you from auto-facing the target which prevents autorot from casting. not really sure how to fix so i'm gonna mark it as wip

BossMod/Autorotation/InputOverride.cs
```diff
@@ -30,6 +30,9 @@ class InputOverride : IDisposable
         private delegate ref int GetRefValueDelegate(int vkCode);
         private GetRefValueDelegate _getKeyRef;
 
+        private IntPtr _forceDisableMovementPtr;
+        internal unsafe ref int ForceDisableMovement => ref *(int*)(_forceDisableMovementPtr + 4);
```
awgil, line 34: Interesting, this seems to be some sort of an internal counter in PlayerController that's used to lock player during animations...

xanunderscore:
> while testing this, I noticed that holding a movement key during movement lock can prevent you from auto-facing the target which prevents autorot from casting. not really sure how to fix so i'm gonna mark it as wip

Ignore this, it looks like it's actually related to the issue I've been having with ReAction, wasn't able to repro it this morning with ReAction disabled.

awgil:
Note to self: I really like this approach way more than messing with input (don't need to care about different keybinds etc), but otoh i really like using LMB+RMB to override movement block. Investigate how to get the best of both worlds...

awgil:
This is no longer relevant after https://github.com/awgil/ffxiv_bossmod/pull/400

PR #218, opened 2024-01-08
SAM rotation

xanunderscore:
basic level 90 samurai rotation. This correctly performs thebalance's standard sam opener, excluding some shuffling of oGCDs which doesn't really matter since samurai is quite flexible.

image

I haven't implemented automatic AI retargeting or auto True North yet.


xanunderscore:
gonna change the branch name lol

PR #221, opened 2024-01-14
Level 90 samurai rotation

xanunderscore:
I haven't tested it enough below level 90 to figure out how to account for the GCD drift introduced by the loss of Ogi, so it's 90 only for now.

### DIFF
--- a/BossMod/Framework/ActionManagerEx.cs
+++ b/BossMod/Framework/ActionManagerEx.cs
@@ -210,6 +210,12 @@ namespace BossMod
             return gcd->Total - gcd->Elapsed;
         }
 
+        public float GCDTime()
+        {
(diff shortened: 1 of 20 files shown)


BossMod/Framework/ActionManagerEx.cs
```diff
@@ -210,6 +210,12 @@ public float GCD()
             return gcd->Total - gcd->Elapsed;
         }
 
+        public float GCDTime()
```
awgil, line 213: What is the intent? If you're trying to get GCD duration - that won't work in downtime.

xanunderscore: That was the intent, yeah. I guess it returns 0 during downtime. I hadn't noticed any issues since all the GCD-length-dependent calculation in the rotation is dependent on being in combat.

PR #224, opened 2024-01-17
Simplified code for Leviathan hard BodySlam AOE

CarnifexOptimus:
realized that I can simplify the code for this a lot. functionality unchanged

### DIFF
--- a/BossMod/BossModule/ArenaColor.cs
+++ b/BossMod/BossModule/ArenaColor.cs
@@ -15,6 +15,5 @@
         public const uint PlayerInteresting = 0xffc0c0c0;
         public const uint PlayerGeneric = 0xff808080;
         public const uint Vulnerable = 0xffff00ff;
-        public const uint Invisible = 0x00FFFFFF;
     }
 }
--- a/BossMod/BossModule/ArenaColor.cs
+++ b/BossMod/BossModule/ArenaColor.cs
@@ -15,5 +15,6 @@
         public const uint PlayerInteresting = 0xffc0c0c0;
         public const uint PlayerGeneric = 0xff808080;
         public const uint Vulnerable = 0xffff00ff;
+        public const uint Invisible = 0x00FFFFFF;
     }
 }
--- a/BossMod/Network/PacketInterceptor.cs
+++ b/BossMod/Network/PacketInterceptor.cs
@@ -42,8 +42,10 @@ namespace BossMod.Network
 
         public unsafe PacketInterceptor()
         {
-            // alternative signature: E8 ?? ?? ?? ?? 84 C0 0F 85 ?? ?? ?? ?? 48 8D 35 - seem to be changing from build to build
-            _fetchHook = Service.Hook.HookFromSignature<FetchReceivedPacketDelegate>("E8 ?? ?? ?? ?? 84 C0 0F 85 ?? ?? ?? ?? 44 0F B6 64 24", FetchReceivedPacketDetour);
+            // alternative signatures - seem to be changing from build to build:
+            // - E8 ?? ?? ?? ?? 84 C0 0F 85 ?? ?? ?? ?? 48 8D 35
+            // - E8 ?? ?? ?? ?? 84 C0 0F 85 ?? ?? ?? ?? 44 0F B6 64 24
+            _fetchHook = Service.Hook.HookFromSignature<FetchReceivedPacketDelegate>("E8 ?? ?? ?? ?? 84 C0 0F 85 ?? ?? ?? ?? 48 8D 35", FetchReceivedPacketDetour);
             Service.Log($"[NPI] FetchReceivedPacket address = 0x{_fetchHook.Address:X}");
 
             // potentially useful sigs from dalamud:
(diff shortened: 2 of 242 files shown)


CarnifexOptimus:
accidently added new module to this pull request, hope it doesnt matter

BossMod/BossModule/ArenaColor.cs
```diff
@@ -15,5 +15,6 @@ public static class ArenaColor
         public const uint PlayerInteresting = 0xffc0c0c0;
         public const uint PlayerGeneric = 0xff808080;
         public const uint Vulnerable = 0xffff00ff;
+        public const uint Invisible = 0x00FFFFFF;
```
awgil, line 18: Why do you need an 'invisible' color, why not just not draw it?

BossMod/Network/PacketInterceptor.cs
```diff
@@ -42,8 +42,10 @@ public bool Active
 
         public unsafe PacketInterceptor()
         {
-            // alternative signature: E8 ?? ?? ?? ?? 84 C0 0F 85 ?? ?? ?? ?? 48 8D 35 - seem to be changing from build to build
-            _fetchHook = Service.Hook.HookFromSignature<FetchReceivedPacketDelegate>("E8 ?? ?? ?? ?? 84 C0 0F 85 ?? ?? ?? ?? 44 0F B6 64 24", FetchReceivedPacketDetour);
+            // alternative signatures - seem to be changing from build to build:
+            // - E8 ?? ?? ?? ?? 84 C0 0F 85 ?? ?? ?? ?? 48 8D 35
+            // - E8 ?? ?? ?? ?? 84 C0 0F 85 ?? ?? ?? ?? 44 0F B6 64 24
+            _fetchHook = Service.Hook.HookFromSignature<FetchReceivedPacketDelegate>("E8 ?? ?? ?? ?? 84 C0 0F 85 ?? ?? ?? ?? 48 8D 35", FetchReceivedPacketDetour);
```
awgil, line 48: Huh? What's this?

CarnifexOptimus: Is it a problem? don't think it will change anything considering it is identical with the existing code

BossMod/BossModule/ArenaColor.cs
CarnifexOptimus: Can I change the genericAOE component for it to
        public override void DrawArenaBackground(BossModule module, int pcSlot, Actor pc, MiniArena arena)
        {
            foreach (var c in ActiveAOEs(module, pcSlot, pc))
                                if (c.Risky)
                c.Shape.Draw(arena, c.Origin, c.Rotation, c.Color);
        }
or are there any instances where you want non risky AOEs to be drawn?

BossMod/Network/PacketInterceptor.cs
awgil: No, don't think so, just wondering...

BossMod/BossModule/ArenaColor.cs
awgil: Yes, the whole idea of Risky is when you want to draw something, but not show warning if player is inside (e.g. if you want to draw beneficial aoes, or if you want to show aoe that will become active much later).

But why would you ever want an invisible AOE? What is it's purpose?
* if it's risky, it would be confusing to player (hint would show 'gtfo', but radar won't show anything)
* if it's non-risky, then it's completely useless? doesn't show hints, doesn't draw aoes...

CarnifexOptimus: the purpose would be to tell the AI that it doesn't have to do anything. Afaik your AI automatically reacts to AOEs even if it is not part of a module (eg the usual trash mobs AOEs). 
So I was thinking if there are fake AOEs to confuse the player or AOEs that only matter if you don't have a specific status effect then the AOE should not be drawn on the radar nor should the AI react to it.

CarnifexOptimus: maybe i can add an optional visibility bool flag to the genericAOEs instead. i will look into that

CarnifexOptimus: ok I pushed an update

awgil: But wait - why do you even report AOE if it's supposed to be ignored?
The intent is for GenericAOE component to report only AOEs that are somewhat relevant to player (either to draw, to show hints, to get out of, etc)

awgil: So if we imagine a usual self-targeted-cast-aoe, that would be ignored for players with a specific status, what you could do is:
```
class MyAOE : Component.SelfTargetedCastAOE
{
  private BitMask _immunePlayers;
  public override OnStatusGain(...) { if ((SID)status.ID == xxx) _immunePlayers.Set(module.Raid.FindSlot(actor.InstanceID)); }
  ... OnStatusLose => similarly clear
  public override ActiveAOEs() { return _immunePlayers[slot] ? Enumerable.Empty<AOEInstance>() : base.ActiveAOEs(...) }
}
```

CarnifexOptimus: because the genericAOE component seems to contain the part that tells AI if a zone is forbidden (if risky), so i was thinking if it is risky=false it will tell the AI that the AOE can be ignored. How else would I tell the AI to ignore a visible AOE?

CarnifexOptimus: ok i removed the fakes from lucky face

PR #225, opened 2024-01-17
added module viewer

Jaksuhn:
adds a modules tab to the vbm window where you can see all available modules.
added parameters you can supply to module info to designate the boss names or content ID which are then used to add more info when the module is registered

new moduleinfo parameters: cfcid (most content), nameid (bnpcname sheet for boss names), notoriousmonsterid (for hunts), fateid (fates), dynamiceventid (critical engagements)

all new parameters are optional and anything uncategorised will be displayed at the bottom of the list using the module's namespace name

most file changes are from adding the relevant moduleinfo parameters to each module

fates are always uncatalogued because there is no way to get the territory type (and thus expansion) via fateID

### DIFF
--- a/BossMod/Config/ConfigUI.cs
+++ b/BossMod/Config/ConfigUI.cs
@@ -33,16 +33,14 @@ namespace BossMod
         private ConfigRoot _root;
         private WorldState _ws;
 
-        private Dictionary<uint, ModuleRegistry.Info> _modules;
-        private Dictionary<uint, ModuleRegistry.Info> _uncatalogued;
-        private Lumina.Excel.ExcelSheet<ExVersion> _exSheet;
-        private Lumina.Excel.ExcelSheet<ContentFinderCondition> _cfcSheet;
-        private List<uint> _expacs;
+        private readonly Lumina.Excel.ExcelSheet<ExVersion> _exSheet;
 
         public ConfigUI(ConfigRoot config, WorldState ws)
         {
             _root = config;
             _ws = ws;
+
+            _exSheet = Service.DataManager.GetExcelSheet<ExVersion>()!;
 
             Dictionary<Type, UINode> nodes = new();
             foreach (var n in config.Nodes)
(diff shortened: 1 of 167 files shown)


BossMod/Config/ConfigUI.cs
```diff
@@ -52,11 +65,80 @@ public ConfigUI(ConfigRoot config, WorldState ws)
...
+            _cfcSheet = Service.DataManager.GetExcelSheet<ContentFinderCondition>()!;
+            _modules = ModuleRegistry.RegisteredModules
+                .Where(x => !x.Value.IsUncatalogued)
+                .OrderBy(x => x.Value.ExVersion)
+                .ThenBy(x => _cfcSheet.GetRow(x.Value.CFCID)?.ClassJobLevelSync)
+                .ThenBy(x => _cfcSheet.GetRow(x.Value.CFCID)?.ItemLevelRequired)
+                .ThenBy(x => _cfcSheet.GetRow(x.Value.CFCID)?.SortKey)
+                .ToDictionary(x => x.Key, x => x.Value);
+            _uncatalogued = _modules.Where(x => x.Value.IsUncatalogued || x.Value.ExVersion == 69).Select(x => x).ToDictionary(x => x.Key, x => x.Value);
+            _expacs = _modules.Where(x => x.Value.ExVersion != 69).Select(x => x.Value.ExVersion).Distinct().ToList()!;
```
awgil, line 79: I think this really belongs to some other class, not config - config should still draw configuration...

BossMod/Debug/DebugObjects.cs
```diff
@@ -41,6 +41,7 @@ public unsafe void DrawObjectTable()
                     _tree.LeafNode($"Gimmick ID: {Utils.ReadField<uint>(internalObj, 0x7C):X}");
                     _tree.LeafNode($"Radius: {obj.HitboxRadius:f3}");
                     _tree.LeafNode($"Owner: {Utils.ObjectString(obj.OwnerId)}");
+                    _tree.LeafNode($"BNpcBase/Name: {obj.DataId}/{((FFXIVClientStructs.FFXIV.Client.Game.Object.GameObject*)obj.Address)->GetNpcID()}");
```
awgil, line 44: Utils.GameObjectInternal(obj)->GetNpcID()

BossMod/Util/UIMisc.cs
```diff
@@ -15,5 +15,17 @@ public static bool DangerousButton(string label)
                 ImGui.SetTooltip("Hold shift");
             return res;
         }
+
+        public static void TextUnderlined(System.Numerics.Vector4 colour, string text)
```
awgil, line 19: Any reason to pass color as vector4 here rather than uint? The rest of the plugin uses uint colors everywhere

Jaksuhn: It was purely because I default to using dalamud colours and didn't notice you don't. I can make a UIColour class or a converter, whichever you'd rather have

BossMod/Config/ConfigUI.cs
Jaksuhn: moved it all to the registry where it makes the most sense at least. Only thing further would be to make a catalogue specific class but that just seems unnecessary

BossMod/Util/UIMisc.cs
Jaksuhn: Oh, also because all Imgui takes in vector4 for text colours

PR #229, opened 2024-01-19
Dancer rotation

xanunderscore:
Tested this in DSR phases 1-3 yesterday, worked as expected. Dancer has no fixed looping rotation due to the randomness of weaponskill combo procs, so the only priorities are using standard and tech step on cooldown (as long as there are enemies in range).

I added a cooldown track to the planner for Standard Step, since it has a 15s duration and you want to be using it as early as possible during downtime - i.e., 14.5 seconds before the boss reappears. I would like to change this to a strategy track at some point so we can disable automatic use of standard step if needed, but since strategy tracks don't show EffectDuration, it's a worse user experience - you have to count for yourself the time elapsed between the start of the Force window and the boss becoming targetable.


Ants1988:
Tried it in P10N and I found that Devilment and Starfall Dance were sometimes getting used outside of Technical finish

Ants1988:
Devilment is being cast after Tilana but is supposed to come immediately after Quadruple Technical Finish

xanunderscore:
> Devilment is being cast after Tilana but is supposed to come immediately after Quadruple Technical Finish

should be fixed by latest commit

Ants1988:
> > Devilment is being cast after Tilana but is supposed to come immediately after Quadruple Technical Finish
> 
> should be fixed by latest commit

Confirmed it fixed the issue.

Thanks!

Ants1988:
I've also noticed the autorotation appears to be using single target spells when multiple enemies are being attacked.

Ants1988:
I've even pressed windmill to get into AOE mode and Im still seeing Cascade, FountainFall and Reverse Cascades.

xanunderscore:
For future reference, Ants1988 was using AI mode with automatic movement disabled, and since I didn't feel like figuring out how to implement target switching, I made AI mode default to single target. However, it's actually not that complicated so now it fully works in AI mode

Ants1988:
Works like a dream now!

PR #232, opened 2024-01-22
Convenience function for finding best AOE target

xanunderscore:


BossMod/Autorotation/RPR/RPRActions.cs
```diff
@@ -44,25 +44,6 @@ public override void Dispose()
 
         public override Targeting SelectBetterTarget(AIHints.Enemy initial)
         {
-            // targeting for aoe
```
awgil, line 47: Is it intended to be fully removed?

xanunderscore: reaper AOE is self targeted so this doesn't do anything. you can see that `NumTargetsHitByAOEGCD` doesn't take a parameter, so the for-loop here is performing the same redundant calculation for every nearby enemy. i think it must have just been copy-pasted from DRG and not updated

PR #235, opened 2024-01-24
Masked Carnivale stage 17-20

CarnifexOptimus:
consider the change to the knockback module for knockback in stage 18 before accepting pull request


BossMod/Modules/MaskedCarnivale/Stage01All'sWellThatStartsWell/Stage01.cs
```diff
@@ -0,0 +1,55 @@
+using System.Linq;
```
awgil, line 1: Again, looks like a duplicate of stuff from other PR?

CarnifexOptimus: you are right
i really need to work on my branch management :(

PR #241, opened 2024-01-26
MNK rotation fixes

xanunderscore:
i skipped over one part of the balance guide's explanation of the braindead looping rotation, and as a result the current version of the MNK rotation is offset from the correct version by one GCD, which causes you to miss out on brotherhood's 5% damage buff for the first Rising Phoenix in your even buff window. this isn't a huge deal but it is suboptimal.

this PR also fixes a mostly cosmetic issue with the positionals predictor. it wouldn't affect True North usage since the positional is only wrong for a brief period during GCD animation lock when you can't use True North anyway, but it is really annoying


xanunderscore:
Optimal drift turned out to not be that hard to implement compared to trying to fix braindead, so I'm just going to open a PR for that

PR #247, opened 2024-01-28
components + sugaar update

CarnifexOptimus:
- Attempt at making two new components (Attracts and simple rotations).
Known issues:
Rotation component will likely not work if multiple actors are doing the rotation (needs to be generalized more)
Attract component will need a better hint to check if destination is unsafe instead of hardcoding something

- Updated Sugaar to use the new components and also fixed the small inaccuracy of using caster.rotation instead of spell.rotation.

### DIFF
--- a/BossMod/Components/Attracts.cs
+++ b/BossMod/Components/Attracts.cs
@@ -1,0 +1,133 @@
+using System;
+using System.Collections.Generic;
+
+namespace BossMod.Components
+{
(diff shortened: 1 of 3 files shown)


awgil:
The new components are good, but I'd like to rewrite them slightly to match the design of other common components. Will do later today.

CarnifexOptimus:
sounds good, feel free to change whatever you like

PR #254, opened 2024-02-01
improvements to Baying of Hellhounds CE

CarnifexOptimus:


BossMod/Modules/Shadowbringers/Foray/CriticalEngagement/CE11ShadowOfDeathHand.cs
```diff
@@ -78,7 +79,27 @@ class BroadsideBarrage : Components.SelfTargetedAOEs
     {
         public BroadsideBarrage() : base(ActionID.MakeSpell(AID.BroadsideBarrage), new AOEShapeRect(40, 20)) { }
     }
+    class BroadsideBarrageKB : Components.Knockback
```
awgil, line 82: What's the point of this module? The point is to avoid AOE, it doesn't matter what exactly it does on failure (kill, apply vuln, knock back into deathwall) - there are tons of AOEs that knockback into deathwall on failure, I tend not to represent that at all.

Knockback components are created for when they are actual mechanics that you need to take into account.

BossMod/Modules/Shadowbringers/Foray/CriticalEngagement/CE12BayingOfHounds.cs
```diff
@@ -47,61 +60,26 @@ class ScorchingLash : Components.SelfTargetedAOEs
...
-
-        private void Activate(WPos source, WPos target, DateTime activation)
-        {
-            var shape = new AOEShapeRect(0, 5);
-            shape.SetEndPoint(target, source, new());
-            _charge = new(shape, source, activation: activation);
-        }
+        public Hellpounce() : base(ActionID.MakeSpell(AID.Hellpounce), 5) { }
+    }
+    class HellpounceSecond : Components.ChargeAOEs
```
awgil, line 67: Ok, so the current implementation would show the second hint for a very short time - this was the reason it was implemented as a custom component before.

Issue #255, opened 2024-02-04
could you add a few options


adambennett55:
could we get a / command to turn on / off AI
like

/vbm aion
/vbm aioff

and could we get a few options added to the boss mod ui under AI to turn Focus Target on and off and follow leader on and off

and could we get a couple / commands to turn these on and off as well like

/vbm fl on
/vbm fl off

and

/vbm ft on
/vbm ft off

and one last thing, could you give a way where other plugins can check to see if Boss Mod is currently Controlling Movement (aka dodging)

McVaxius:
someone seems to have done some of that already in a pull request:

https://github.com/awgil/ffxiv_bossmod/pull/251

adambennett55:
I also implemented most of this, minus the event. #256

awgil:
Assumed to be fixed, since relevant PR is now merged.

PR #256, opened 2024-02-08
Most of the requested items from "could you add a few options #255"

adambennett55:
Only haven't added the event to trigger when vbm is controlling movement.

### DIFF
--- a/BossMod/AI/AIBehaviour.cs
+++ b/BossMod/AI/AIBehaviour.cs
@@ -36,7 +36,7 @@ namespace BossMod.AI
                 return;
 
             // keep master in focus
-            FocusMaster(master);
+            //FocusMaster(master);
 
             _afkMode = !master.InCombat && (_autorot.WorldState.CurrentTime - _masterLastMoved).TotalSeconds > 10;
             bool forbidActions = _forbidActions || _ctrl.IsMounted || _afkMode || _autorot.ClassActions == null || _autorot.ClassActions.AutoAction >= CommonActions.AutoActionFirstCustom;
--- a/BossMod/AI/AIBehaviour.cs
+++ b/BossMod/AI/AIBehaviour.cs
@@ -36,7 +36,8 @@ namespace BossMod.AI
                 return;
 
             // keep master in focus
-            //FocusMaster(master);
+            if (Plugin._ai._config.FocusTargetLeader)
+                FocusMaster(master);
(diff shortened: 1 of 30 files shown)


BossMod/Framework/Plugin.cs
```diff
@@ -46,8 +47,8 @@ public sealed class Plugin : IDalamudPlugin
         private WorldStateGame _ws;
         private BossModuleManager _bossmod;
         private Autorotation _autorotation;
-        private AI.AIManager _ai;
-        private AI.Broadcast _broadcast;
+        public static AIManager _ai;
```
awgil, line 50: Please no :(

BossMod/Framework/Plugin.cs
```diff
@@ -161,6 +162,38 @@ private void OnCommand(string cmd, string args)
                 case "r":
                     _wndReplay.SetVisible(!_wndReplay.IsOpen);
                     break;
+                case "aion":
```
awgil, line 165: Why is that needed, considering there's already a 'cfg' command for modifying arbitrary options?

BossMod/AI/AIBehaviour.cs
```diff
@@ -36,7 +36,8 @@ public void Execute(Actor player, Actor master)
                 return;
 
             // keep master in focus
-            FocusMaster(master);
+            if (Plugin._ai._config.FocusTargetLeader)
```
awgil, line 39: Config is accessible directly via Service.Config.Get if needed, was this the only reason for making the _ai field static?

BossMod/AI/AIManager.cs
```diff
@@ -7,11 +7,11 @@
 
 namespace BossMod.AI
 {
-    class AIManager : IDisposable
+    public class AIManager : IDisposable
     {
-        private Autorotation _autorot;
+        public Autorotation _autorot;
```
awgil, line 12: The plugin follows standard c# naming conventions (private fields _likeThis, public LikeThis). See https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names

awgil:
> Sorry I'll fix it

Btw you don't have to close/recreate PR to update it :) just push changes and it will be updated automatically!

adambennett55:
It's been updated sorry for the trouble

PR #258, opened 2024-02-09
module for PvP Hidden Gorge - Goblin Mercenary

CarnifexOptimus:
possibly incomplete, since i was unable to get a long replay, but this is better than nothing.


BossMod/Modules/MaskedCarnivale/Stage04GentlemanPreferSwords/Stage04Act2.cs
```diff
@@ -80,7 +80,6 @@ protected override void DrawEnemies(int pcSlot, Actor pc)
 
         public override void CalculateAIHints(int slot, Actor actor, PartyRolesConfig.Assignment assignment, AIHints hints)
         {
-            base.CalculateAIHints(slot, actor, assignment, hints);
```
awgil, line 83: Are you 100% sure? This isn't a very good idea, since the base class implementation does important things (assign bounds and gathers hints from all components).

Since this PR is about PVP boss, I'm going to assume this is unintended and revert - please open another PR if you really want to do that (and I'd be interested to hear the rationale).

CarnifexOptimus: I am not sure actually, I just removed it because i saw no difference after removing it

PR #262, opened 2024-02-11
Add outlines to actor and several components

CarnifexOptimus:
Proposing to add very thin outlines to actor and several component mechanics to increase visibility depending on background (mainly Color.Danger).


awgil:
Eh, i dunno, not sure it will always look good, will think about it...

CarnifexOptimus:
i haven't seen any case yet where it looks bad (i guess good/bad look is subjective anyway), but I guess I can't rule out that situations like that might arrise. Still beats changing colours on a case by case basis in my opinion
image
image2
the change is really subtle but makes a difference imo

CarnifexOptimus:
Added a toggle to the outlines/shadows so people have a choice

PR #264, opened 2024-02-12
added anti knockback abilities

CarnifexOptimus:
Added 3 anti knockback abilities to the knockback component. I hope I have done this right and also not forget about any.

### DIFF
--- a/BossMod/Components/Knockback.cs
+++ b/BossMod/Components/Knockback.cs
@@ -129,7 +129,7 @@ namespace BossMod.Components
         }
 
         protected IEnumerable<(WPos from, WPos to)> CalculateMovements(BossModule module, int slot, Actor actor)
-        {
+        {   
             if (MaxCasts <= 0)
                 yield break;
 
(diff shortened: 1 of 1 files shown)


BossMod/Components/Knockback.cs
```diff
@@ -30,6 +30,7 @@ protected struct PlayerImmuneState
         {
             public DateTime ArmsLengthSurecastExpire; // 0 if not active
             public DateTime InnerStrengthExpire; // 0 if not active
+            public DateTime ManawallExpire; // 0 if not active
 
             public bool ImmuneAt(DateTime time) => ArmsLengthSurecastExpire > time || InnerStrengthExpire > time;
```
awgil, line 35: You also need to include the check here for it to actually work.

BossMod/Components/Knockback.cs
```diff
@@ -30,6 +30,7 @@ protected struct PlayerImmuneState
         {
             public DateTime ArmsLengthSurecastExpire; // 0 if not active
             public DateTime InnerStrengthExpire; // 0 if not active
+            public DateTime ManawallExpire; // 0 if not active
```
awgil, line 33: Maybe rename to Role/Job/Duty?

CarnifexOptimus: good idea

BossMod/Components/Knockback.cs
CarnifexOptimus: fixed it

PR #265, opened 2024-02-13
Updated Gurangatch module to use the new rotation component

CarnifexOptimus:
I updated the Gurangatch module to use the new rotation component. I had to set sequence index to 0 because otherwise the rotation sometimes broke.
Also set the index on Sugaar to 0 as a safety precaution.

### DIFF
--- a/BossMod/Modules/Endwalker/HuntA/Gurangatch.cs
+++ b/BossMod/Modules/Endwalker/HuntA/Gurangatch.cs
@@ -37,7 +37,7 @@ namespace BossMod.Endwalker.HuntA.Gurangatch
             {
                 AID.OctupleSlammerLCW or AID.OctupleSlammerRCW => 90.Degrees(),
                 AID.OctupleSlammerLCCW or AID.OctupleSlammerRCCW => -90.Degrees(),
-                AID.LeftHammerSecond or AID.RightHammerSlammer => 180.Degrees(),
+                AID.LeftHammerSlammer or AID.RightHammerSlammer => 180.Degrees(),
                 _ => default
             };
             if (increment != default)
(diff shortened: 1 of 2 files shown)


BossMod/Modules/Endwalker/HuntA/Gurangatch.cs
```diff
@@ -25,84 +25,44 @@ public enum AID : uint
         BoneShaker = 27512, // Boss->self, 4.0s cast, range 30 circle
     };
 
-    class Slammer : Components.GenericAOEs
+    class Slammer : Components.GenericRotatingAOE
     {
-        private int _remainingSlams = 0;
-        private Angle _slamDir;
-        private Angle _slamDirIncrement;
+        private Angle _increment;
```
awgil, line 30: Do you actually need this field here? Looking at the code, it's always set in OnCastStart, then reset back immediately after - could just use a local variable?

BossMod/Modules/Shadowbringers/HuntA/Sugaar.cs
```diff
@@ -88,7 +88,7 @@ public override void OnCastStarted(BossModule module, Actor caster, ActorCastInf
         public override void OnCastFinished(BossModule module, Actor caster, ActorCastInfo spell)
         {
             if ((AID)spell.Action.ID is AID.NumbingNoiseRotating or AID.TailSnapRotating)
-                AdvanceSequence(caster.Position, spell.Rotation, module.WorldState.CurrentTime);
+                AdvanceSequence(0, module.WorldState.CurrentTime);
```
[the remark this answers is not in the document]
CarnifexOptimus: I do not have a replay at this point, i just added it as a safety precaution since sequences breaking at gurangatch and acheloios (even if replay was not provided by the discord user) were observed. Since there can only be one sugaar at a time I guess it does not have any downsides?
I will add the sequences count

BossMod/Modules/Endwalker/HuntA/Gurangatch.cs
CarnifexOptimus: updated it

PR #266, opened 2024-02-13
Updated Burfurlur rotation for new component

CarnifexOptimus:


BossMod/Modules/Endwalker/HuntS/Armstrong.cs
```diff
@@ -22,50 +23,86 @@ public enum AID : uint
         SoporificGas = 27478, // Boss->self, 6.0s cast, range 12 circle
     };
 
-    class MagitekCompressor : Components.GenericAOEs
+    class MagitekCompressor : Components.GenericRotatingAOE
```
awgil, line 26: This one seems to be overcomplicated - surely you can just invert the sequence's increment after 5th cast?

BossMod/Modules/Endwalker/HuntS/Burfurlur.cs
```diff
@@ -11,66 +11,61 @@ public enum AID : uint
...
+        QuintupleSneeze1 = 27309, // Boss->self, 5.0s cast, range 40 45-degree cone
+        QuintupleSneeze24 = 27310, // Boss->self, 0.5s cast, range 40 45-degree cone
+        QuintupleInhale35 = 27692, // Boss->self, 0.5s cast, range 27 45-degree cone
+        QuintupleSneeze35 = 27693, // Boss->self, 0.5s cast, range 40 45-degree cone
         Uppercut = 27314, // Boss->self, 3.0s cast, range 15 120-degree cone
         RottenSpores = 27313, // Boss->location, 3.0s cast, range 6 circle
     };
 
-    class QuintupleSneeze : Components.GenericAOEs
+    class QuintupleSneeze : Components.GenericRotatingAOE
```
awgil, line 24: It's only barely a 'rotation' component (the sequence has very arbitrary increments), don't making it a rotation actually simplifies anything...

BossMod/Modules/Endwalker/HuntS/Armstrong.cs
CarnifexOptimus: i am not sure if I understand what you mean, but the activation delay is different (3.6s instead of 2.1s) for the cast where it is inverting other than that already is inverting the increment after the 5th cast

BossMod/Modules/Endwalker/HuntS/Burfurlur.cs
CarnifexOptimus: I don't know if it is now more complicated than before, but at least the activation times seem to be perfect. And the colors are now consistant with other mechanics now where sequences exist

PR #271, opened 2024-02-21
Support for queueing and executing ExecuteCommand

xanunderscore:
ExecuteCommand is required for NIN (`/statusoff`) and for swapping Lost Actions in Bozja, which is part of the optimal DRS opener for every class

### DIFF
--- a/BossMod/Autorotation/CommonActions.cs
+++ b/BossMod/Autorotation/CommonActions.cs
@@ -21,6 +21,8 @@ namespace BossMod
             public Actor? Target;
             public Vector3 TargetPos;
             public ActionSource Source;
+            public CommandID Command;
+            public (uint, uint, uint, int) Arguments;
 
             public NextAction(ActionID action, Actor? target, Vector3 targetPos, ActionSource source)
             {
@@ -28,6 +30,14 @@ namespace BossMod
                 Target = target;
                 TargetPos = targetPos;
                 Source = source;
+            }
+
+            public static NextAction ExecuteCommand(CommandID id, uint arg1, uint arg2, uint arg3, int arg4)
+            {
+                return new NextAction() {
+                    Command = id, 
+                    Arguments = (arg1, arg2, arg3, arg4)
+                };
             }
         }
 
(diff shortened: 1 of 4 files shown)


awgil:
Eh, I don't really think this is the best approach. The autorotation module abstracts away game's implementation detail, and this solution requires modules to know implementation details (command ids and args).

Suggestion:
1. for statusoff - this isn't really an action (it can be executed inside animlock etc); what exactly are the usecases?
1.1. iirc there was something for NIN (cancelling stealth or w/e), could be just handled by Actions, since it's not really part of rotation
1.2. shit like removing KB immune via planner could be useful. Would probably be some part of AIHints structure? (statuses to remove)
2. UseLostAction - now this is rotational ability; there's an actual ActionID for it. Maybe the rotation could simply return desired AID, and then framework would check whether it is currently set into holster, and if not - do the execute command?
2.1. this won't really solve the 'force resocket action to reset cooldown' usecase though
2.2. should it really be used from action manager's update?..

I'd like to investigate this more myself before making any decision.

xanunderscore:
>     1. for statusoff - this isn't really an action (it can be executed inside animlock etc); what exactly are the usecases?

the only one that's actually a required part of a rotation is `/statusoff Hidden` for NIN, at least until square enix adds an Unhide action.

FWIW, i am doing it in NINActions currently, and not in the rotation code. I used to just call ExecuteCommand directly, but it felt weird just doing that in one specific rotation's update loop. like, at least ActionManagerEx is logically an incoming queue that processes action requests. maybe you have a better solution in mind?

> UseLostAction - now this is rotational ability; there's an actual ActionID for it. Maybe the rotation could simply return desired AID, and then framework would check whether it is currently set into holster, and if not - do the execute command?

this sounds good except for the case where the rotation returns an ActionID for a lost action that the user isn't carrying. this would result in a situation kind of like what you get in a buggy/unfinished rotation where you forget to add an `Unlocked()` check and the rotation keeps returning the same (unusable) action indefinitely.

awgil:
Ok, I think latest commits provide all the necessary functionality.

PR #274, opened 2024-02-23
basic ipc

Jaksuhn:
mainly meant for visland and SND integration

The only problem I have is that InitiateCombat will only trigger the first 2-3 actions before it will always have the NextAction be None. Also not sure if I should be using `AutoActionFirstCustom` or `AutoActionAIFight`


Jaksuhn:
closing in favour of #272 that I did not see existed. I'll just add onto that

PR #275, opened 2024-02-23
Swap ArenaColor to ComponentType enum

sourpuh:
This is a precursor change for supporting configurable colors. This is mostly a 1:1 swap from a user facing perspective, but there are a handful of minor changes to tether colors in P4S, P7S, Zurvan Unreal, and possibly a couple more.


sourpuh:
Need to rethink this

PR #276, opened 2024-02-25
Expanded BaitAwayCast

CarnifexOptimus:
Added bools for charges and endoncastevent (only used on Sugriva for now)

### DIFF
--- a/BossMod/Components/BaitAway.cs
+++ b/BossMod/Components/BaitAway.cs
@@ -178,22 +178,45 @@ namespace BossMod.Components
     public class BaitAwayCast : GenericBaitAway
     {
         public AOEShape Shape;
+        public bool ResolvesAtCastEvent;
+        public bool IsCharge;
+        public float HalfWidthCharge;
 
-        public BaitAwayCast(ActionID aid, AOEShape shape, bool centerAtTarget = false) : base(aid, centerAtTarget: centerAtTarget)
+        public BaitAwayCast(ActionID aid, AOEShape shape, bool centerAtTarget = false, bool isCharge = false, float halfWidthCharge = 0, bool resolvessAtCastEvent = false) : base(aid, centerAtTarget: centerAtTarget)
         {
             Shape = shape;
+            IsCharge = isCharge;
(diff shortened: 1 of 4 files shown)


BossMod/Components/BaitAway.cs
```diff
@@ -178,22 +178,45 @@ public override void OnUntethered(BossModule module, Actor source, ActorTetherIn
     public class BaitAwayCast : GenericBaitAway
     {
         public AOEShape Shape;
+        public bool ResolvesAtCastEvent;
```
awgil, line 181: Not a fan of this one in the generic components, it's too easy to misuse by mistake, without understanding all implications. There are good reasons why cast-start should generally be paired with cast-finish - each cast-start is _guaranteed_ to have a corresponding cast-finish (eg if spell is interrupted or canceled, if actor is destroyed, etc - and actor destruction can happen for a number of reasons), and obviously there is no such guarantee for cast-event.

Are there any examples other than sugriva that need this? If not, I'd prefer if it stayed specific to that component, with the comment describing why it is done the way it is done, and only promoted to generic component once we have evidence that this is a mechanic of the game (and then maybe something similar would be needed for other casts...)

BossMod/Components/BaitAway.cs
```diff
@@ -178,22 +178,45 @@ public override void OnUntethered(BossModule module, Actor source, ActorTetherIn
...
+        public bool IsCharge;
+        public float HalfWidthCharge;
```
awgil, line 183: This should be a separate component (eg without Shape field). Otherwise - what's the purpose of defining a Shape for a charge? The constructor would still require it, etc.

BossMod/Components/BaitAway.cs
CarnifexOptimus: well there are still 100s of encounters without a module and I doubt that they made this mechanic just for one open world boss. but i understand your point.

BossMod/Components/BaitAway.cs
CarnifexOptimus: ok, i will try to rework it later into a seperate component

PR #277, opened 2024-02-25
Even more treasure hunter modules

CarnifexOptimus:


BossMod/Modules/PVP/HiddenGorge/GoblinMercenary.cs
```diff
@@ -130,7 +130,7 @@ public override void OnCastStarted(BossModule module, Actor caster, ActorCastInf
         {
             if ((AID)spell.Action.ID == AID.GobfireShootypopsStart)
             {
-                Sequences.Add(new(_shape, caster.Position, spell.Rotation, _increment, spell.NPCFinishAt, 1, 6));
+                Sequences.Add(new(_shape, caster.Position, spell.Rotation, _increment, spell.NPCFinishAt.AddMilliseconds(-52), 1, 6));
```
awgil, line 133: Don't use precision finer than 0.1s (or at very least 0.05s) - this simply makes no sense, due to ping and client and server framerates - that's just noise.

In this particular case, just using NPCFinishAt is simpler and good enough.

CarnifexOptimus: ok

CarnifexOptimus: i changed it, i think it doesn't matter anyway since using AI in pvp is pointless right, so this was just for future proofing

PR #281, opened 2024-02-26
Draw no active AOE for generic raidwides

xanunderscore:


BossMod/Autorotation/AutoHints.cs
```diff
@@ -7,6 +7,8 @@ namespace BossMod
     // this is used e.g. in outdoor or on trash, where we have no active bossmodules
     public class AutoHints : IDisposable
     {
+        static readonly float RAIDWIDE_SIZE = 40;
```
awgil, line 10: VBM uses standard C# naming conventions: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names

Specifically:
* Use PascalCase for constant names, both fields and local constants.

BossMod/Autorotation/AutoHints.cs
```diff
@@ -50,10 +52,11 @@ private void OnCastStarted(object? sender, Actor actor)
                 return;
             AOEShape? shape = data.CastType switch
             {
-                2 => new AOEShapeCircle(data.EffectRange), // used for some point-blank aoes and enemy location-targeted - does not add caster hitbox
+                // used for some point-blank aoes and enemy location-targeted - does not add caster hitbox
+                2 => data.EffectRange >= RAIDWIDE_SIZE ? null : new AOEShapeCircle(data.EffectRange),
```
awgil, line 56: null return here will trigger error message later. I'd just check this above and return (eg if EffectRange >= const && data.CastType is 2 or 5)

CarnifexOptimus:
Just a note that this change can lead to problems with gazes and AOEs with dmg fall off since they usually have the same raidwide size. Should be judged carefully if the trade off is worth it.

PR #286, opened 2024-03-01
Additions to two zadnor CE modules

CarnifexOptimus:
Additions to two zadnor CE modules
also attempted to add AI to baitawaychargecast, looks fine in AI hints during replays, but not sure how it turns out in reality (maybe choppy movement?)


CarnifexOptimus:
added a CastInterruptHint component for good measure

CarnifexOptimus:
added Knockback AI to Never Cry Wolf CE, as requested by LTS

BossMod/Components/CastHint.cs
```diff
@@ -3,6 +3,19 @@
 
 namespace BossMod.Components
 {
+    public enum AID : uint
```
awgil, line 6: This will pollute very common namespace (Components) with a very common type name (AID) that only actually contains very specific action id's relevant only to one component (and even then - only for AI of all things). This should either be a private subtype of interrupt hint component, or moved to proper location (eg. Autorotation/BLU/...).

Also, this should probably be moved to a different PR, as it's not really related to zadnor CE's, as PR name implies.

BossMod/Components/CastHint.cs
```diff
@@ -36,4 +49,78 @@ public override void OnCastFinished(BossModule module, Actor caster, ActorCastIn
...
+        }
+
+        public override void AddAIHints(BossModule module, int slot, Actor actor, PartyRolesConfig.Assignment assignment, AIHints hints)
+        {
+            base.AddAIHints(module, slot, actor, assignment, hints);
+            foreach (var caster in _casters)
+            {
+                if (Active && Canbeinterrupted && actor.Role == Role.Tank && Service.ClientState.LocalPlayer?.Level >= 18)
+                    hints.PlannedActions.Add((ActionID.MakeSpell(WAR.AID.Interject), caster, 1, false));
+                if (Active && Canbeinterrupted && actor.Class.GetClassCategory() == ClassCategory.PhysRanged && Service.ClientState.LocalPlayer?.Level >= 24)
```
awgil, line 99: You can't access ClientState from components, because it won't be available in UIDev. Everything that is needed for components should be part of the world state.

awgil: Ok, I've added Level field to Actor.

BossMod/Components/CastHint.cs
```diff
@@ -36,4 +49,78 @@ public override void OnCastFinished(BossModule module, Actor caster, ActorCastIn
                 _casters.Remove(caster);
         }
     }
+
+    public class CastInterruptHint : CastCounter
```
awgil, line 53: Why isn't it a subclass of CastHint?

BossMod/Components/LineOfSightAOE.cs
```diff
@@ -73,7 +73,7 @@ public override void AddAIHints(BossModule module, int slot, Actor actor, PartyR
             if (BlockersImpassable)
             {
                 var blockers = Blockers.Select(b => ShapeDistance.Circle(b.Center, b.Radius)).ToArray();
-                hints.AddForbiddenZone(p => blockers.Min(b => b(p)));
+                hints.AddForbiddenZone(p => blockers.Min(b => b(p)), NextExplosion);
```
awgil, line 76: This isn't right. The idea is that AOE happens in a few seconds, and blockers are already there - you have to move around them to reach safespot.

BossMod/Components/LineOfSightAOE.cs
```diff
@@ -110,6 +110,7 @@ public override void OnCastStarted(BossModule module, Actor caster, ActorCastInf
             {
                 _casters.Add(caster);
                 Refresh(module);
+                NextExplosion = spell.NPCFinishAt;
```
awgil, line 113: This isn't right - this should be done by Refresh(), using active caster's NPCFinishAt.

BossMod/Components/CastHint.cs
awgil: In fact, this is wrong - it should not hardcode all possible interrupt actions - rather, it should set appropriate Enemy.ShouldBeInterrupted (and new ShouldBeStunned) flag, and let autorotation modules handle it.

BossMod/Components/CastHint.cs
CarnifexOptimus: I will see what I can do. Sadly, noone made anything for BLU autorotation yet, so I had this idea since I have no plans to do any work on autorotations.

BossMod/Components/CastHint.cs
CarnifexOptimus: I can change it, but then every autorotation needs support for shouldbeinterrupted/stunned I guess

BossMod/Components/CastHint.cs
CarnifexOptimus: I can do that, but then there needs to be an additional Hint string in the constructor it seems

BossMod/Components/LineOfSightAOE.cs
CarnifexOptimus: I guess I will remove it for now. I don't really understand why this would be needed since it seems to work properly when looking at AI hints in replay.

BossMod/Components/CastHint.cs
awgil: Yep, and that's exactly what they should do

PR #295, opened 2024-03-12
Basic Zermous Normal support

LTS-FFXIV:
mostly reusing code from extreme, as a way to teach myself modules work

TODO: some random stuff isnt appearing like blackhole from the left which is odd


awgil:
This looks like a copy paste of extreme, down to states and timings - is normal really so identical to the extreme?

LTS-FFXIV:
extreme zermous is just normal zeromus that hits harder and has i think 2 new mechanics, timings, ranges, etc are all the same between fights

BossMod/Modules/Endwalker/NormalTrials/Trial7Zeromus/Trial7ZeromusStates.cs
```diff
@@ -0,0 +1,302 @@
...
+        private void AbyssalNoxEchoesSableThread(uint id, float delay, bool second)
+        {
+            Cast(id, AID.AbyssalNox, delay, 5);
+            ComponentCondition<AbyssalEchoes>(id + 0x1000, 0.1f, comp => comp.Casters.Count > 0)
+                .ActivateOnEnter<AbyssalEchoes>()
+                .ExecOnEnter<AbyssalEchoes>(comp => comp.MaxCasts = 0); // don't show aoes before doom, this is misleading
+            ComponentCondition<AbyssalEchoes>(id + 0x1010, 5, comp => comp.Casters.Count > 5, "1 hp"); // dooms are slightly staggered apply around here
+            ComponentCondition<AbyssalEchoes>(id + 0x1020, 11, comp => comp.NumCasts > 0, "Circles 1")
+                .ExecOnEnter<AbyssalEchoes>(comp => comp.MaxCasts = 5);
+            ComponentCondition<AbyssalEchoes>(id + 0x1030, 5, comp => comp.Casters.Count == 0, "Circles 2")
```
awgil, line 57: Ok, I've just watched a video of a normal version of the trial - for this very first mechanic I don't see the two staggered sets of aoes with 5s between them, like extreme has.

So I'm confused why is this a thing here.

BossMod/Modules/Endwalker/NormalTrials/Trial7Zeromus/Trial7ZeromusStates.cs
```diff
@@ -0,0 +1,302 @@
...
+        private State VisceralWhirl(uint id, float delay)
+        {
+            CastMulti(id, new[] { AID.VisceralWhirlR, AID.VisceralWhirlL }, delay, 8)
+                .ActivateOnEnter<VisceralWhirl>();
+            ComponentCondition<VisceralWhirl>(id + 2, 0.8f, comp => !comp.Active, "Lines")
+                .DeactivateOnExit<VisceralWhirl>();
+            ComponentCondition<MiasmicBlast>(id + 0x10, 0.3f, comp => comp.Casters.Count > 0)
+                .ActivateOnEnter<MiasmicBlast>();
+            return ComponentCondition<MiasmicBlast>(id + 0x11, 8, comp => comp.NumCasts > 0, "Crosses")
+                .DeactivateOnExit<MiasmicBlast>();
```
awgil, line 109: I don't see this bit after visceral whirl on normal mode either...

awgil:
> extreme zermous is just normal zeromus that hits harder and has i think 2 new mechanics, timings, ranges, etc are all the same between fights

Sorry, I don't believe that's true - I've just checked the video and this is not correct for the first two mechanics, they have way more things going on in the extreme than in normal. I did not check the rest. From my experience, different flavours (normal/extreme trials, normal/savage raids, etc) all have some similarity with the abilities, but the details vary a lot, to the point where it's not reasonable to try copy-pasting extreme to make normal work. Some components would probably be similar, but the states would be different for sure.

LTS-FFXIV:
I just ran this activity myself twice, once with the state machine as it was, copy pasted, and once with a simple state.
A few things. 

1, I'd like to emphasize that this is in fact my first eve

r module i created, before I had a full grasp of how the statemachine worked and how you specifically tend to utilize it. 

2, the fights are in fact almost the exact same, and with a simple state just loading the components, the module would work if the components had proper removes for their aoes and werent entirely reliant on the statemachine. 

Ill be swapping the state machine to a simplestate and fixing the individual components so their drawings disappear at the appropriate time

below is a preview as is with simplestate machine
https://github.com/awgil/ffxiv_bossmod/assets/127939494/43db5049-71c3-4685-994b-f7af919debdd

PR #296, opened 2024-03-12
Basic GATE support

LTS-FFXIV:
Done to teach myself how to make modules from scratch, not incredibly useful at the moment aside from dodging Daigoro in slice is right

TODO: tie stuff to animations where possible instead, especially in slice is right

PR #300, opened 2024-03-14
The Slice is right module

CarnifexOptimus:
Since the modules from pull request #296 are unfinished I tried my hand at making better. Here is the Slice is Right module, it contains everything except the solution to the cup phase which seems to be fully server sided, but I will keep investigating.
I don't think you can make an "anyway the wind blows" module, all attacks seem to be server side until the player cannot move anymore, so a module seems pointless there. In best case you could probably suggest a spot where you are least likely to be hit by any attacks.


BossMod/Modules/GoldSaucer/TheSliceIsRight/TheSliceIsRight.cs
```diff
@@ -0,0 +1,160 @@
...
+            if ((OID)actor.OID == OID.Bamboo && !_bamboospawn.Any(b => b.Position.AlmostEqual(actor.Position, 1)))
+                _bamboospawn.Add(actor);
+        }
+
+        public override void Update(BossModule module)
+        {
+            if (_time != default && module.WorldState.CurrentTime > _time)
+            {
+                _time = default;
+                _circle = _circle.Except(_circleToberemoved).ToList();
```
awgil, line 82: This looks pointlessly roundabout and inefficient - why not do something that would convey the intent better, e. g.
```
_circle.RemoveAll(c => _circleToBeRemoved.Contains(c))
```

BossMod/Modules/GoldSaucer/TheSliceIsRight/TheSliceIsRight.cs
```diff
@@ -0,0 +1,160 @@
...
+                _circleToberemoved.Clear();
+                _singlesplit = _singlesplit.Except(_singlesplitToberemoved).ToList();
+                _singlesplitToberemoved.Clear();
+                _doublesidedsplit = _doublesidedsplit.Except(_doublesidedsplitToberemoved).ToList();
+                _doublesidedsplit.Clear();
```
awgil, line 87: And this is just pointless - you first rebuild a list, and then immediately clear it?..

BossMod/Modules/GoldSaucer/TheSliceIsRight/TheSliceIsRight.cs
CarnifexOptimus: ok i changed it, hope it doesn't change the functionality

PR #301, opened 2024-03-16
Hydaelyn nm module

CarnifexOptimus:


BossMod/Modules/Endwalker/Trials/T02Hydaelyn/Echoes.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        public int NumCasts { get; private set; }
+
+        public Echoes() : base(6, 0, 8) { }
+
+        public override void OnEventCast(BossModule module, Actor caster, ActorCastEvent spell)
+        {
+            if ((AID)spell.Action.ID == AID.Echoes)
+                ++NumCasts;
+                if (NumCasts == 5)
```
awgil, line 13: This is extremely misleading. It is indented, as if it were part of condition above, but in reality it is not.

Please use auto formatting.

BossMod/Modules/Endwalker/Trials/T02Hydaelyn/Exodus.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+
+namespace BossMod.Endwalker.Trials.T02Hydaelyn
+{
+   class Exodus : BossComponent
+    {
+        private int crystalsdestroyed;
+        private bool casting;
```
awgil, line 8: This is not a cast? This looks like something that should be implemented via state machine.

BossMod/Modules/Endwalker/Trials/T02Hydaelyn/WeaponTracker.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                staff = true;
+                chakram = false;
+                sword = false;
+            }
+            if ((SID)status.ID == SID.HydaelynsWeapon && status.Extra == 0x1B5)
+            {
+                staff = false;
+                chakram = true;
+                sword = false;
```
awgil, line 44: Should be an enum instead?

BossMod/Modules/Endwalker/Trials/T02Hydaelyn/Echoes.cs
CarnifexOptimus: it was working as intended, but sure i can add another set of brackets to it

BossMod/Modules/Endwalker/Trials/T02Hydaelyn/Exodus.cs
CarnifexOptimus: Exodus is a cast, it says cast event in the raw ops
I don't understand how the statemachine works so i am trying to avoid using more than trivial phase, plus the statemachine doesn't add a blue hint or predicted damage

awgil: Blue hint is redundant for raidwides covered by statemachines.

Predicted damage is done by checking next state with raidwide hint.

CarnifexOptimus: noone is looking at the ugly wall of white text in the state machine though, maybe if it would be presented better like in cactbot it would be worth the effort learning it

PR #302, opened 2024-03-16
added AOEShapeTriangle

LTS-FFXIV:
created triangle shape to handle Thaliaks abilities, can likely be revised to be better but functions similarly to AOEShapeCone but this is the best i can do


LTS-FFXIV:
note: this is for equilateral triangles not isosceles

awgil:
I'll wait until your full implementation of Thaliak module and think about what to do with it.

This triangle shape seems to be very unique, used by a single fight. I'm not quite sure whether it makes sense to have it among generic shapes, since there is no single obviously correct definition for a triangle shape - e.g. your shape has origin at one of the vertices (like cone or rect), but for equilateral triangle it might just as well be a center (like your triangle arena border!). It's also not clear why generic 'triangle aoe shape' should be limited to equilateral triangles.

LTS-FFXIV:
regarding the shape origin point, this is basically AOEShapeCone. much like how that ones origin point is its vertex, the only abiltiy in the game so far to use triangles is a self cast directional AOE so having it cast from one of its verticies is correct. if it was cast from its center, much like AOEShapeCircle, it would be wrong

its the difference between this
image

and this (ignore orientation, this one has the origin point at the center, which is wrong)
image

BossMod/BossModule/AOEShapes.cs
```diff
@@ -234,4 +234,61 @@ private IEnumerable<WPos> ContourPoints(WPos origin, Angle rotation, float offse
...
+        public override IEnumerable<IEnumerable<WPos>> Contour(WPos origin, Angle rotation, float offset = 0, float maxError = 1)
+        {
+            var vertices = CalculateVertices(origin, rotation + DirectionOffset, offset);
+            return new List<IEnumerable<WPos>> { new[] { vertices.p1, vertices.p2, vertices.p3 } };
+        }
+
+        public override Func<WPos, float> Distance(WPos origin, Angle rotation)
+        {
+            // Implementing an exact distance calculation for a triangle shape might be complex and is beyond the scope of this basic implementation.
+            return p => (p - origin).Length(); // Simplified placeholder
```
awgil, line 276: This is just an intersection (== max) of 3 half-planes. Very similar to ShapeDistance.Rect, with different optimization opportunities.

PR #303, opened 2024-03-16
Added ArenaBoundsTri

LTS-FFXIV:
Basic triangle shaped arena for Thaliak mechanics, centered on coordinate point, can likely be revised to be better, arena size value (HalfSize) used represents the radius of a circumscribed circle around the equilateral triangle to determine size


BossMod/BossModule/ArenaBounds.cs
```diff
@@ -209,4 +209,69 @@ public override WDir ClampToBounds(WDir offset, float scale)
...
+        {
+            var a = Center + new WDir(-HalfSize, HalfSize * sqrt3 / 3);
+            var b = Center + new WDir(HalfSize, HalfSize * sqrt3 / 3);
+            var c = Center + new WDir(0, -2 * HalfSize * sqrt3 / 3);
+        
+            bool b1 = Sign(p, a, b) < 0.0f;
+            bool b2 = Sign(p, b, c) < 0.0f;
+            bool b3 = Sign(p, c, a) < 0.0f;
+        
+            return ((b1 == b2) && (b2 == b3));
```
awgil, line 247: This is just `p.InTri(a, b, c)`?

BossMod/BossModule/ArenaBounds.cs
```diff
@@ -209,4 +209,69 @@ public override WDir ClampToBounds(WDir offset, float scale)
...
+            yield return Center + new WDir(-halfSide, height / 3);
+            yield return Center + new WDir(halfSide, height / 3);
+            yield return Center + new WDir(0, -2 * height / 3);
+        }
+
+        public override Pathfinding.Map BuildMap(float resolution = 0.5f)
+        {
+            // BuildMap implementation for equilateral triangle
+            // This is a simplified example and would need to be adapted based on specific pathfinding requirements
+            throw new NotImplementedException();
```
awgil, line 234: Should be very similar to circle - create a map and block out stuff outside triangle. Keeping it like that and throwing an exception will cause issues (eg with replay viewer).

BossMod/BossModule/ArenaBounds.cs
```diff
@@ -209,4 +209,69 @@ public override WDir ClampToBounds(WDir offset, float scale)
...
+        {
+            // Define triangle vertices
+            var a = Center + new WDir(-HalfSize, HalfSize * sqrt3 / 3);
+            var b = Center + new WDir(HalfSize, HalfSize * sqrt3 / 3);
+            var c = Center + new WDir(0, -2 * HalfSize * sqrt3 / 3);
+        
+            // Ray-triangle intersection algorithm goes here
+            // This is a complex topic and requires a bit of math
+            // Placeholder for the actual intersection calculation
+            return float.NaN; // Return NaN to indicate that this method needs proper implementation
```
awgil, line 266: Should use a new helper, similar to Intersect.RayQuad - Intersect.RayTri should simply check 3 segments instead of 4.

BossMod/BossModule/ArenaBounds.cs
```diff
@@ -209,4 +209,69 @@ public override WDir ClampToBounds(WDir offset, float scale)
...
+        }
+
+
+        public override WDir ClampToBounds(WDir offset, float scale = 1)
+        {
+            // Clamping within a triangle is highly context-dependent
+            // This method needs a detailed implementation based on specific requirements
+            return new WDir(0, 0); // Placeholder to indicate that clamping logic is needed
```
awgil, line 274: This will force drawing actors outside bounds at center, which would be quite ugly. The implementation is effectively finding an intersection with a ray cast from arena center with arena bounds.

awgil:
I suggest combining this and the aoeshapetriange with the implementation of the module using them (thaliak) - i'll then review and clean that up at the same time, otherwise it's hard for me to debug that.

PR #308, opened 2024-03-19
Reworked interrupt hint

CarnifexOptimus:
reworked interrupt hint to make showing a name optional


awgil:
All of that code just for the name? A lot of that is duplicated.

I think it would be much simpler to add a name as an explicit hint, where it makes sense (eg if you have several mobs active and it's important who to interrupt). I know that localization could be an issue, but tbh I don't really care, as the plugin doesn't handle localization properly anyway, and I'd prefer consistency to having only small bits localized here and there.

CarnifexOptimus:
Ok I reworked the hint again, it is much shorter now
maybe we should look into adding more localisation options instead of not doing any

PR #309, opened 2024-03-19
PersistentInvertibleVoidzone component

CarnifexOptimus:
For croizat's module.

Other usecases: Masked Carnivale Stage 28, Shinryu, Heroe's Gauntlet, Forgall in City of Mhach and who knows what other places use this

### DIFF
--- a/BossMod/Components/PersistentVoidzone.cs
+++ b/BossMod/Components/PersistentVoidzone.cs
@@ -1,7 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.Drawing;
 using System.Linq;
 
 namespace BossMod.Components
--- a/BossMod/Components/PersistentVoidzone.cs
+++ b/BossMod/Components/PersistentVoidzone.cs
@@ -1,5 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Drawing;
+using System.Linq;
 
 namespace BossMod.Components
 {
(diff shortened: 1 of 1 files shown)


awgil:
I don't quite understand the intent of this component.

The normal voidzone is simply a persistent aoe effect that is presumed to be forbidden to enter (e.g. immediately applying a dot or otherwise killing player when he enters it).

What are 'invertible' voidzones?
# i understand you typically have to stand inside them when some cast finishes
# what happens if you stay inside them when cast is _not_ being cast? is it safe/required/irrelevant? From the name, it seems to act as a usual voidzone?
# what happens if there are multiple 'inverted zones'? i presume you need to stand in any of them?

CarnifexOptimus:
> I don't quite understand the intent of this component.
> 
> The normal voidzone is simply a persistent aoe effect that is presumed to be forbidden to enter (e.g. immediately applying a dot or otherwise killing player when he enters it).
> 
> What are 'invertible' voidzones?
> 
>     1. i understand you typically have to stand inside them when some cast finishes
> 
>     2. what happens if you stay inside them when cast is _not_ being cast? is it safe/required/irrelevant? From the name, it seems to act as a usual voidzone?
> 
>     3. what happens if there are multiple 'inverted zones'? i presume you need to stand in any of them?
> 
>     4. can some of them be inverted while others are not?

2. They usually give a status effect while standing inside, that is either beneficial or detrimental depending on the current cast. Normally this it ends in an instant kill if you are inside or outside at the wrong time. 
In Croizat's module case simply nothing happens if you stand in the puddle while neither cast is ending, but in other cases like the Heroe's Gauntlet dungeon you only want to stand a limited time in the voidzone, because a detrimental effect will activate if you stand inside for too long (usually about 5s limit).
3. If there are multiple you can stand in any of them.
4. I havent seen a case of this yet, so they are all or nothing if the OID stays the same for now, but who knows what the future will bring.

PR #311, opened 2024-03-19
module viewer updates + endwalker solo duty module

Jaksuhn:
requires #309 to be merged for the endwalker module

module viewer is now a filterable table with added support for quest battles, as well as name and exversion overrides for content where that's missing (unreals, fates)

I also added the proper moduleinfo for any modules that were uncatalogued


McVaxius:
image
for situations like this.  is it possible to have like a "Unknown" Category so it can be unselected

PR #315, opened 2024-03-22
added auto recording of replays

Jaksuhn:
options to auto start/stop on duty start/end

changed replay files to be prefixed by the duty name (or zone name if that doesn't exist) instead of "World"
It also adds the appends the duty finder settings (unrestricted, level sync, min ilvl, no echo) as comma separated values after the instance name so `the Bowl of Embers_U,NE_2024_03_22_20_18_27.log` as an example

### DIFF
--- a/BossMod/Config/ConfigNode.cs
+++ b/BossMod/Config/ConfigNode.cs
@@ -60,6 +60,22 @@ namespace BossMod
         }
     }
 
+    // attribute that specifies slider should be used for displaying float property
+    [AttributeUsage(AttributeTargets.Field)]
+    public class PropertyIntSliderAttribute : Attribute
+    {
+        public float Speed = 1;
+        public int Min;
+        public int Max;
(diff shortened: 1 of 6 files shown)


McVaxius:
is it possible to disaggregate (just use random number for uuid + text for name)( player identifiable info for sharing replays?

McVaxius:
merged and tested this PR this morning seems to be working fine for me

McVaxius:
the final msq solo duty didnt auto stop recording when i left it also
(leaving when it was concluded in this case)

Jaksuhn:
> found a bug with it
> 
> ran a sunken temple of qarn (normal one), then 20 minutes later queued into a zeromus and both got recorded. under the one qarn recording.
> 
> upon reviewing the replay. it seems like it never stopped recording.
> 
> i suspect its because i left the qarn 2/3 of the way through instead of completing it.

addressed

PR #328, opened 2024-04-04
Thaleia

CarnifexOptimus:
This is LTS work I am reworked because there were many issues.
All bosses + trash packs done pending minor additions at Eulogia to Quintessence

Things that could be done:
- Rewriting triangle border and shape code as they look AI generated
- Investigating why knockback from source is direction 6 now
- consider adding support for concave polygon arena shapes
- making timelines


CarnifexOptimus:
seems like i am finished for now except maybe some minor additions to Quintessence at Eulogia, found 41 ENVC/AID combos in 14 logs and probably would find more if I had more logs

PR #330, opened 2024-04-04
Add IPC's:

adambennett55:
ActiveModuleComponentBaseList
ActiveModuleComponentList
ActiveModuleHasComponent

### DIFF
--- a/BossMod/Framework/IPCProvider.cs
+++ b/BossMod/Framework/IPCProvider.cs
@@ -1,4 +1,5 @@
-﻿using Dalamud.Game.ClientState.Objects.Types;
+﻿using BossMod.Components;
+using Dalamud.Game.ClientState.Objects.Types;
 
 namespace BossMod;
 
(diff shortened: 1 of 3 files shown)


BossMod/Framework/IPCProvider.cs
```diff
@@ -1,13 +1,17 @@
...
 class IPCProvider : IDisposable
 {
     private List<Action> _disposeActions = new();
 
-    public IPCProvider(Autorotation autorotation)
+    public IPCProvider(Autorotation autorotation, BossModuleManager bossmod)
     {
+        Register("ActiveModuleComponentBaseList", () => bossmod.ActiveModule?.Components.Select(c => c.GetType().BaseType?.Name).ToList() ?? default);
+        Register("ActiveModuleComponentList", () => bossmod.ActiveModule?.Components.Select(c => c.GetType().Name).ToList() ?? default);
+        Register("ActiveModuleHasComponent", (string name) => bossmod.ActiveModule?.Components.Any(c => c.GetType().Name == name || c.GetType().BaseType?.Name == name) ?? false);
```
awgil, line 14: Why would you need these IPCs even?

adambennett55:
For AutoDuty, Specifically I check if the ActiveModule has a cleave Component, in which case I ignore that specific ForbiddenZone in order to determine if i am allowed to move my character. because when there is forbidden zones i stop all movement and let vbm take over

awgil:
Uh, that sounds extremely fragile. Can you explain why exactly would you want to do custom movement in combat (irrespective of whether cleave is active)? Presumably, when in combat, you just want to do your dps rotation (and stay in place or move for positionals)?

adambennett55:
> Uh, that sounds extremely fragile. Can you explain why exactly would you want to do custom movement in combat (irrespective of whether cleave is active)? Presumably, when in combat, you just want to do your dps rotation (and stay in place or move for positionals)?
> 
> I don't mind merging it as is really, but maybe there's a better way...

just in situations where dodging gets you out of range or on the edge of the arena, it likes to do that alot, in those situations as tank ill move closer to the middle of the arena or if dps either move closer to mob or healer if need heals or if healer closer to tank

awgil:
Right, ok, this should really be handled by ai itself. But I guess since I'm not really improving it soon, I'll just merge this and let you deal with that on your side.

PR #334, opened 2024-04-06
fix for replay filenames

Jaksuhn:
forgot that some duties have illegal characters in them which results in a failure to create a replay file for those instances

### DIFF
--- a/BossMod/Replay/ReplayManagementWindow.cs
+++ b/BossMod/Replay/ReplayManagementWindow.cs
@@ -171,6 +171,6 @@ public class ReplayManagementWindow : UIWindow
         if (cf->IsSilenceEcho)
             prefix += "_NE";
 
-        return prefix;
+        return Utils.CleanFilename(prefix);
     }
 }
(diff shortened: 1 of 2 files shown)


BossMod/Replay/ReplayManagementWindow.cs
```diff
@@ -171,6 +171,6 @@ private unsafe string GetPrefix()
         if (cf->IsSilenceEcho)
             prefix += "_NE";
 
-        return prefix;
+        return Utils.CleanFilename(prefix);
```
awgil, line 174: Uhh why exactly is it needed? There's a call to StringToIdentifier above already, which removes all non-alphanumeric characters - what exactly is the part of the name that creates invalid characters?

Jaksuhn: my fault. This branch was from before the last changes to auto record but I didn't PR it then cause I couldn't test it. Tested it now then remerged master but didn't see your changes already fixed this too

PR #335, opened 2024-04-07
Thaleia optimisations, Azeyma Radiant Flourish change, Nophica Matrons Breath change

CarnifexOptimus:
still thinking about updating the rest of azeyma solar fans, but I dont have a log where it only hits 6 instead of 8 times at the moment


CarnifexOptimus:
closing because trying to solve merge conflicts messed everything up

PR #344, opened 2024-04-16
ArenaBoundsPolygon, ArenaBoundsUnion, ArenaBoundsDifference, ArenaBoundsDonut, improvement to alliance raids, Masked Carnivale 28-32, caching, other small changes 

CarnifexOptimus:
Known issue with polygon arenas: - none, should work for any non-self-intersecting polygon without holes
Known issues with union arenas: - doesn't build correctly if there is more than one non-intersecting hole or more than 2 disjointed shapes

I added Quadtrees and Voronoi Diagrams while testing various algorithms for the polygon arenas, but they are not used in the current implementation. So if you don't have any use for the stuff there, consider deleting those two cs files.


CarnifexOptimus:
Redacting this pull requests since I disagree with any licensing agreement you are trying to add. GTFO

PR #350, opened 2024-04-30
TOP p1-p3 adjustments for aether strats

xanunderscore:
NA raid community aka Aether generally uses Sausage Roll's strats for TOP

P1 toolbox: https://ff14.toolboxgaming.space/?id=758088204654761&preview=1
P2 toolbox: https://ff14.toolboxgaming.space/?id=043186470764761&preview=1
P3 monitors raidplan: https://raidplan.io/plan/om8w5ea548ajBE45

differences from existing module:

1. Group numbers for P1 are inverted - G1 is NW CCW, G2 is N CW - I didn't change this because I figured it would be too much trouble for a purely aesthetic change, the existing module does priority fine
2. Pivot point for Pantokrator safe spots is NE/SW instead of N/S
3. Party Synergy G2 order for Remote Glitch is GPOB - purple and orange don't swap, they just go to arena edge
4. Swap priority for P2 flare stacks is north rather than south
5. 3rd monitor positions vertically south, instead of 1st monitor positioning vertically north - prio works the same way

There are presumably additional differences in later phases but I haven't gotten that far since I prog in PF

### DIFF
--- a/BossMod/Modules/Endwalker/Ultimate/TOP/P1Pantokrator.cs
+++ b/BossMod/Modules/Endwalker/Ultimate/TOP/P1Pantokrator.cs
@@ -25,7 +25,13 @@ class P1FlameThrower(BossModule module) : Components.GenericAOEs(module)
         var group = _pantokrator != null ? _pantokrator.PlayerStates[pcSlot].Group : 0;
         if (group > 0)
         {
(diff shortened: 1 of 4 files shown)


xanunderscore:
whoops gotta fix my shitty math

xanunderscore:
the logic for panto NE/SW is completely wrong, not sure what i did but this needs another pass

xanunderscore:
ok, it wasn't THAT wrong

PR #355, opened 2024-05-12
Orthos Floor 90 Boss Complete/Floor 80 Boss Started

LeontopodiumNivale14:
Completion of the floor 90 boss in Eureka Orthos DD, started the base layout for floor 80, you can ignore it entirely if possible for now. Might work on that more tomorrow

### DIFF
--- a/UIDev/UITest.cs
+++ b/UIDev/UITest.cs
@@ -146,8 +146,27 @@ class UITest
                         }
                     }
                 }
+
+                // Should return "C:\Program Files (x86)\Steam\steamapps\common\game Online" if installed with default options.
+                Span<int> validSteamAppIds = [39210, 312060];
+                foreach (var steamAppId in validSteamAppIds)
+                {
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@ obj/
 bin/
 *.user
 logs/
+UIDev/UITest.cs
```
awgil, line 6: Local change that doesn't belong to this PR?

UIDev/UITest.cs
```diff
@@ -148,6 +148,6 @@ private static string FindGameDataPath()
                 }
             }
         }
-        return "D:\\installed\\SquareEnix\\FINAL FANTASY XIV - A Realm Reborn\\game\\sqpack";
+        return "C:\\M.2 Steam Library\\steamapps\\common\\FINAL FANTASY XIV Online\\game\\sqpack";
```
awgil, line 151: Local change that doesn't belong to this PR?

BTW I've now grabbed more recent version of this code from xivlauncher that accounts for steam version, so this should no longer be needed for you.

PR #381, opened 2024-07-11
DT Dancer

xanunderscore:
the rotation barely changed. tillana has a slightly different and more annoying behavior but i can't be bothered to go back into the lab to fully optimize it...maybe when savage comes out...

very little cdplan support atm, just a boolean flag for whether to use tech step or not


BossMod/BossModule/AIHints.cs
```diff
@@ -64,6 +68,7 @@ public class Enemy(Actor actor, bool shouldBeTanked)
     // clear all stored data
     public void Clear()
     {
+        ForceMovementIn = float.MaxValue;
```
xanunderscore, line 71: i need to fix this, this is the wrong place for it. it tends to stick when you disable ai mode

PR #382, opened 2024-07-12
DT dungeon fixes

xanunderscore:
in addition to the fixes outlined below, i uncommented the line in AIBehaviour that resets the desired camera rotation after gaze mechanics since D022 has one of those. according to blame, that line was last changed 2 years ago, so if there is a good reason it was commented out please let me know

D013:
1. voidzone from 8-way lightning AOEs lingered too long, i forgot to check eventstate
2. donut AOE after spread tankbuster was too large
3. 8-way wind AOEs have only two patterns for a total of 6 possible spawn points, so hardcode instead of doing the neat fade effect
4. i forgot the voidzone for the whirlwind itself whoopsie

D021:
1. i just thought this status wasn't ever used but it is, i can't read

D022:
1. just assumed the size of this donut aoe without cross referencing it with footage


CarnifexOptimus:
seems like you are learning the powers of copy & pase

PR #391, opened 2024-07-21
Add MCH class utility

Aimsucks:
- Added MCH class utility for Tactician and Dismantle tracks


Aimsucks:
No idea why this removed the latest commit. Will address and make a new PR.

PR #401, opened 2024-07-27
some rotations

xanunderscore:
VERY basic rotations for 
- gunbreaker, paladin
- sage
- monk, ninja, samurai, viper, ~~dragoon~~ (i haven't gotten it to level 100 yet)
- black mage, picto, summoner

i am sorry for being this cringe, but i thought opening a separate PR for every class would actually be worse (because then it would be 11 separate PRs, but they also all depend on the changes i've made to my common base class, so it's a lose lose situation - third party rotation providers will save us)

just getting these out there so we can get more people using vbm and so i can get more feedback!


BossMod/ActionQueue/ActionDefinition.cs
```diff
@@ -86,7 +100,8 @@ public float ChargeCapIn(ReadOnlySpan<Cooldown> cooldowns, int level)
         if (MainCooldownGroup < 0)
             return 0;
         var cdg = cooldowns[MainCooldownGroup];
-        return cdg.Total > 0 ? (MaxChargesAtLevel(level) * Cooldown - cdg.Elapsed) : 0;
+        // TODO MaxChargesAtLevel is zero for some actions with charges (like Air Anchor), but cdg.Total can only be nonzero if the action is on cooldown, i.e. if it was used, i.e. it has more than 0 charges
+        return cdg.Total > 0 ? (Math.Max(1, MaxChargesAtLevel(level)) * Cooldown - cdg.Elapsed) : 0;
```
awgil, line 104: I think MaxChargesAtLevel shouldn't ever return 0... Is that because sheets have 0? Probably better to clamp it to 1 there.

xanunderscore: it's 0 for probably 99% of actions, like GCDs that don't have a charge system, but it's *also* zero for single-charge GCDs that also have their own separate cooldown. one example is machinist Air Anchor. does it make sense for regular GCDs to default to having 1 max charge?

awgil: Yeah i think so, normally action has one charge.

PR #408, opened 2024-08-04
rotation updates

xanunderscore:
I got DRG to 100 so now it's `Basic` instead of `WIP`

savage patch removed the only moderately interesting feature of VPR and broke the rotation as a result, added fix here

brand new shiny RPR rotation! it tries to do double shroud. will need further testing

### DIFF
--- a/BossMod/BossModule/RaidCooldowns.cs
+++ b/BossMod/BossModule/RaidCooldowns.cs
@@ -38,9 +38,19 @@ public sealed class RaidCooldowns : IDisposable
         return MathF.Max(0, (float)(firstAvailable - _ws.CurrentTime).TotalSeconds);
     }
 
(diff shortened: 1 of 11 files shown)


xanunderscore:
the source of the build error is my WIP M01S module that isn't in this PR...so...thanks github

BossMod/BossModule/RaidCooldowns.cs
```diff
@@ -38,9 +38,19 @@ public float NextDamageBuffIn()
         return MathF.Max(0, (float)(firstAvailable - _ws.CurrentTime).TotalSeconds);
     }
 
+    public float NextDamageBuffIn2()
```
awgil, line 41: Huh what's this?

PR #412, opened 2024-08-10
dtr bar display

xanunderscore:
With icon: 
image

without: 
image

umbra version: 
image

popup: 
image

the popup is kind of small but it is the exact same scale as the regular autorotation window - i wonder if it would be a good idea to make it bigger...

### DIFF
--- a/BossMod/Framework/Plugin.cs
+++ b/BossMod/Framework/Plugin.cs
@@ -82,7 +82,7 @@ public sealed class Plugin : IDalamudPlugin
         _ai = new(_rotation, _amex);
         _broadcast = new();
         _ipc = new(_rotation, _amex);
-        _dtr = new(_rotation);
+        _dtr = new(_rotation, _ai);
 
         Service.Framework.Update += OnUpdate;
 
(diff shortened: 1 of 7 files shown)


adambennett55:
Looks nice I'll remove the strbat from my pull In lieu of yours

BossMod/Framework/Plugin.cs
```diff
@@ -81,6 +82,9 @@ public unsafe Plugin(IDalamudPluginInterface dalamud, ICommandManager commandMan
         _ai = new(_rotation, _amex);
         _broadcast = new();
         _ipc = new(_rotation, _amex);
+        _dtr = new(_rotation);
+
+        Service.Framework.Update += OnUpdate;
```
awgil, line 87: Why do you need it to happen on framework update, rather than on usual plugin update (which happens during draw)?

xanunderscore: oh no idea I copied this from vnavmesh, i can move it

awgil: When you update it, pls merge to master, i've recently merged other PR which added some dtr stuff, I think it's duplicated now.

xanunderscore:
Updated, I moved herc's DTR bar into this class

PR #413, opened 2024-08-10
another big PR

xanunderscore:
### new rotations
- RDM
- WHM
- BRD
- healer AI

### rotation changes
- tank AI will use (more?) mits
- ranged AI will limit break if using duty support (lmao)
- helper functions for picking dot targets. still needs work

### framework changes
- add ClassJobLevels to ClientState (for role action unlock check). this makes #410 redundant
- allow modules to pick their own preferred target/range for AI mode. this was needed for RDM since it has a melee combo but is otherwise a ranged dps.
- add `TransformPosition` to action definitions, which was needed for AST's earthly star as it cannot target enemies
- use the same ActionUnlocked check in rotations and in manual queue override


BossMod/Data/PartyState.cs
```diff
@@ -45,9 +45,9 @@ void assign(ulong instanceID, Actor? actor)
     }
 
     // select non-null and optionally alive raid members
-    public IEnumerable<Actor> WithoutSlot(bool includeDead = false)
+    public IEnumerable<Actor> WithoutSlot(bool includeDead = false, bool partyOnly = true)
```
awgil, line 48: This is a big breaking change, why do you need it?

BossMod/ActionQueue/ActionDefinition.cs
```diff
@@ -53,11 +54,13 @@ public sealed record class ActionDefinition(ActionID ID)
     public ActionAspect Aspect; // useful for BLM and BLU
     public int MaxChargesBase; // baseline max-charges when action is unlocked
     public readonly List<(int Charges, int Level, uint UnlockLink)> MaxChargesOverride = []; // trait overrides for max-charges (applied in order)
+    public bool IsRoleAction; // unlocked conditions are different for these
     public float InstantAnimLock = 0.6f; // animation lock if ability is instant-cast
     public float CastAnimLock = 0.1f; // animation lock if ability is non-instant
     public ConditionDelegate? ForbidExecute; // optional condition, if it returns true, action is not executed
     public SmartTargetDelegate? SmartTarget; // optional target transformation for 'smart targeting' feature
     public TransformAngleDelegate? TransformAngle; // optional facing angle transformation
+    public TransformPositionDelegate? TransformPosition; // optional ground-targeted action position transformation
```
awgil, line 63: Ok this one I don't quite understand, what exactly are you trying to achieve?

BossMod/Autorotation/RotationModule.cs
```diff
@@ -93,12 +93,9 @@ public abstract class RotationModule(RotationModuleManager manager, Actor player
 
     public virtual string DescribeState() => "";
 
-    // utility to check action/trait unlocks
-    public bool ActionUnlocked(ActionID action)
-    {
-        var def = ActionDefinitions.Instance[action];
-        return def != null && def.AllowedClasses[(int)Player.Class] && Player.Level >= def.MinLevel && (ActionDefinitions.Instance.UnlockCheck?.Invoke(def.UnlockLink) ?? true);
-    }
+    public virtual AI.Targeting? SelectTargetForAI(StrategyValues strategy, AI.Targeting aiDecision) => null;
```
awgil, line 96: Why is this needed?

BossMod/Data/PartyState.cs
xanunderscore: oh right, good catch. i'll change the default value to be false

BossMod/Autorotation/RotationModule.cs
xanunderscore: just for range to target. EW bossmod had similar functionality. (specifically, red mage wants to be around 25y from target in most cases, and in melee range if they have enough gauge.)

BossMod/ActionQueue/ActionDefinition.cs
xanunderscore: for ground targeting, specifically for actions that can't target enemies - AST Earthly Star, whatever SCH and WHM's ground targeted heals/mits are called (and caster LBs if we ever want to cdplan those)

PR #414, opened 2024-08-11
more rotations

xanunderscore:
- DRK
- AST
- SCH

rotation fixes
- mnk auto-SSS before downtime
- rdm AI range-to-target calculation
- healer AI raise fix (lol)


xanunderscore:
don't forget to make the `targetPos` change

PR #418, opened 2024-08-14
add default presets

xanunderscore:


awgil:
Will think about it. I think we need some sort of a FTUE welcome screen, that would allow setting up sample presets...

PR #420, opened 2024-08-14
ipc methods for preset creation

xanunderscore:


adambennett55:
Could you possibly also add an IPC for setting the current preset for current class and getting the current preset for the current class

awgil:
> Could you possibly also add an IPC for setting the current preset for current class and getting the current preset for the current class

I'm curious - why do you want different presets for different classes? The intention is to have 'situational' presets (eg single-target, aoe, etc), each of which has modules for all classes you care about.

adambennett55:
> > Could you possibly also add an IPC for setting the current preset for current class and getting the current preset for the current class
> 
> I'm curious - why do you want different presets for different classes? The intention is to have 'situational' presets (eg single-target, aoe, etc), each of which has modules for all classes you care about.

well its not for different presets for different classes, AD would use the same preset for all but we just need the ability to set our preset as the AI preset, and to check if its already set to that, to turn Autorot on

awgil:
Ah, so we need some extra IPC to modify AI preset then? And you can call this ipc after (re)importing AD preset.

adambennett55:
yes exactly

awgil:
Added AI.SetPreset IPC

PR #429, opened 2024-08-20
ClassPLDUtility.cs update; ClassDRKUtility.cs implementation

Akechi-kun:
- [x] Add _Stance_ to all **Tank Utility** modules
- [x] Test _Stance_ & rest of general **Tank** actions listed in **Utility** module
- [x] Add appropriate actions related to **PLD/DRK**
- [x] Add _CDPlanner_ support for **PLD/DRK**
- [x] Test _CDPlanner_ support for **PLD/DRK**


BossMod/Autorotation/Utility/RoleTankUtility.cs
```diff
@@ -29,7 +29,7 @@ protected static void DefineShared(RotationModuleDefinition def, ActionID lb3, A
         DefineSimpleConfig(def, SharedTrack.Shirk, "Shirk", "", 150, ClassShared.AID.Shirk);
         DefineSimpleConfig(def, SharedTrack.ArmsLength, "ArmsLength", "ArmsL", 300, ClassShared.AID.ArmsLength, 6); // note: secondary effect 15s
 
-        def.Define(SharedTrack.Stance).As<StanceOption>("Stance", "", -10)
+        def.Define(SharedTrack.Stance).As<StanceOption>("Stance", "", 10)
```
awgil, line 32: The reason these things were negative is to hide them by default, since they are so rarely useful (i.e. interject is needed for like 1 fight). Stance is probably more useful for offtanks, so I guess making it positive is reasonable.

However, I've just realized that 'usual' (non-replay-based) UI for cdplan has no way of toggling column visibility! Will fix.

PR #435, opened 2024-08-22
AI mode: check priority only

xanunderscore:
this changes AI mode to default to attacking *any* priority target it can get its hands on. this would allow useful features like IPC methods to set specific mob priorities (by instanceid) so that AI mode can be instructed to fight them. it's also a lot easier to read than the previous conditional, OUCHIE

obviously we don't want to constantly try to fight every aggroed mob in the current zone, so i have also modified `FillPotentialTargets` to consider in-combat targets as priority only if they are attacking a party member (or alliance member)

note that this won't work in large scale content like dalriada or DRN where the boss you're fighting might be targeting a tank from another alliance, but anyone who uses full hands-off AI mode in bozja probably doesn't give a shit anyway.

### DIFF
--- a/BossMod/BossModule/AIHints.cs
+++ b/BossMod/BossModule/AIHints.cs
@@ -10,7 +10,7 @@ public sealed class AIHints
         public const int PriorityForbidFully = -2; // attacking this enemy is forbidden both by ai or player (e.g. invulnerable, or attacking/killing might lead to a wipe)
 
         public Actor Actor = actor;
-        public int Priority = PriorityForbidAI; // <0 means damaging is actually forbidden, 0 is default (TODO: revise default...)
+        public int Priority = actor.InCombat ? 0 : PriorityForbidAI; // <0 means damaging is actually forbidden, 0 is default (TODO: revise default...)
         //public float TimeToKill;
         public float AttackStrength = 0.05f; // target's predicted HP percent is decreased by this amount (0.05 by default)
         public WPos DesiredPosition = actor.Position; // tank AI will try to move enemy to this position
--- a/BossMod/BossModule/AIHints.cs
+++ b/BossMod/BossModule/AIHints.cs
@@ -10,7 +10,7 @@ public sealed class AIHints
         public const int PriorityForbidFully = -2; // attacking this enemy is forbidden both by ai or player (e.g. invulnerable, or attacking/killing might lead to a wipe)
 
         public Actor Actor = actor;
-        public int Priority = actor.InCombat || actor.FateID != 0 ? 0 : PriorityForbidAI; // <0 means damaging is actually forbidden, 0 is default (TODO: revise default...)
+        public int Priority = PriorityForbidAI; // <0 means damaging is actually forbidden, 0 is default (TODO: revise default...)
         //public float TimeToKill;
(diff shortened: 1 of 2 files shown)


BossMod/BossModule/AIHints.cs
```diff
@@ -10,7 +10,7 @@ public class Enemy(Actor actor, bool shouldBeTanked)
         public const int PriorityForbidFully = -2; // attacking this enemy is forbidden both by ai or player (e.g. invulnerable, or attacking/killing might lead to a wipe)
 
         public Actor Actor = actor;
-        public int Priority = actor.InCombat || actor.FateID != 0 ? 0 : PriorityForbidAI; // <0 means damaging is actually forbidden, 0 is default (TODO: revise default...)
+        public int Priority = PriorityForbidAI; // <0 means damaging is actually forbidden, 0 is default (TODO: revise default...)
```
awgil, line 13: There are a bunch of Enemy instances created outside of FillPotentialTargets (i.e. arr titan modules), they probably need to be updated...

xanunderscore: no i can just revert this, since i'm always setting priority in FillPotentialTargets anyway

BossMod/BossModule/AIHints.cs
```diff
@@ -86,9 +86,19 @@ public void FillPotentialTargets(WorldState ws, bool playerIsDefaultTank)
...
-        foreach (var actor in ws.Actors.Where(a => a.Type == ActorType.Enemy && a.IsTargetable && !a.IsAlly && !a.IsDead && (a.FateID == 0 || a.FateID == allowedFateID)))
+        foreach (var actor in ws.Actors.Where(a => a.Type == ActorType.Enemy && a.IsTargetable && !a.IsAlly && !a.IsDead))
         {
-            PotentialTargets.Add(new(actor, playerIsDefaultTank && (actor.InCombat || actor.FateID != 0)));
+            // fate mob in fate we are NOT a part of, skip entirely. it's okay to "attack" these (i.e., they won't be added as forbidden targets) because we can't even hit them
+            // (though aggro'd mobs will continue attacking us after we unsync, but who really cares)
+            if (actor.FateID > 0 && actor.FateID != allowedFateID)
+                continue;
+
+            var allowedAttack = actor.InCombat && ws.Party.FindSlot(actor.TargetID) >= 0;
```
awgil, line 96: So uhh, if i pull a mob and no one targets it, i'm fucked? I don't really follow, why is it needed?

xanunderscore: if you pull a mob and no one targets it, then it would be targeting you, so nothing would really change. this is just to prevent ai mode from attacking random other players' targets in overworld

awgil: Ah right i see, ok

xanunderscore:
one unexpected consequence of this change is that it also kinda interferes with solo duties where you have a lot of npc allies...i wonder if it would be better to additionally check whether some mob is in the enmity list

xanunderscore:
#444 makes this PR obsolete if you prefer that approach, FYI

awgil:
Do i need to merge this before #444?

PR #451, opened 2024-08-28
AI Hints for ChainsComponent

VeraNala:
- chainLength is the minimum distance players must be apart to break the chain, while initalDistance considers starting distance between the pair
- Additionally added an optional param for rare (I think only Ifrit atm, maybe?) chains that require the pair to stay together, rather than break. (Maybe O12S-P2 Blue/R&G tethers if this component is used for that)
- In that case the minimum distance should represent the distance the pair can move without triggering the chain.

- Reuploaded to address EOL issues (I hope? Sorry Veyn)

### DIFF
--- a/BossMod/Components/Chains.cs
+++ b/BossMod/Components/Chains.cs
@@ -1,11 +1,12 @@
 ﻿namespace BossMod.Components;
 
-// component for breakable chains
+// component for breakable chains - Note that chainLength for AI considers the minimum distance needed for a chain-pair to be broken (assuming perfectly stacked at cast)
 public class Chains(BossModule module, uint tetherID, ActionID aid = default) : CastCounter(module, aid)
 {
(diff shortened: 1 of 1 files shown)


BossMod/Components/Chains.cs
```diff
@@ -34,6 +45,7 @@ public override void OnTethered(Actor source, ActorTetherInfo tether)
             {
                 SetPartner(source.InstanceID, target);
                 SetPartner(target.InstanceID, source);
+                _initialDistance = (source.Position - target.Position).Length();
```
xanunderscore, line 48: this should probably be per-partner, like the `_partner` array is. you could even make that field a `(Actor? Partner, float InitialDistance)[]` instead. otherwise, if 4 people get chained at once, the initial distance will be effectively random, it'll just get set to whoever received the chain event last

PR #453, opened 2024-08-28
Keeper of the Lake Modules

LeontopodiumNivale14:
-> Boss 1 & 2 are complete
-> Boss 3 is 98% Complete, the Ahk Morn module isn't disappearing after the 4th hit (3rd hit from the follow up, 4th if you include the initial)

I'll poke into this more, but most groups shouldn't even see the full 1st stack (it's going to bother me regardless)

### DIFF
--- a/BossMod/Modules/RealmReborn/Dungeon/D18TheKeeperoftheLake/D183Midgardsormr.cs
+++ b/BossMod/Modules/RealmReborn/Dungeon/D18TheKeeperoftheLake/D183Midgardsormr.cs
@@ -1,4 +1,6 @@
-﻿namespace BossMod.Modules.RealmReborn.Dungeon.D18TheKeeperoftheLake.D183Midgardsormr;
+﻿using System.ComponentModel;
+
+namespace BossMod.Modules.RealmReborn.Dungeon.D18TheKeeperoftheLake.D183Midgardsormr;
 
 public enum OID : uint
 {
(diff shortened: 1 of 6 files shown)


BossMod/Modules/RealmReborn/Dungeon/D18TheKeeperoftheLake/D183Midgardsormr.cs
```diff
@@ -0,0 +1,125 @@
...
+class AhkMornFollow(BossModule module) : Components.UniformStackSpread(module, 6, 0, 4)
+{
+    public int NumCasts { get; private set; }
+
+    public override void OnEventCast(Actor caster, ActorCastEvent spell)
+    {
+        if ((AID)spell.Action.ID is AID.AkhMornFollowup or AID.AkhMornCast)
+            ++NumCasts;
+
+        if ((AID)spell.Action.ID is AID.AkhMornFollowup or AID.AkhMornCast && NumCasts >= 4)
```
xanunderscore, line 87: you should probably check `NumCasts >= 3` if you're having issues with the AOE not disappearing. if akh morn is 4 total hits, that means only 3 of them are the follow up action, so this will never reach 4 anyway

BossMod/Modules/RealmReborn/Dungeon/D18TheKeeperoftheLake/D183Midgardsormr.cs
```diff
@@ -0,0 +1,125 @@
...
+            AID.AntipathyCast2 => 1,
+            AID.AntipathyCast3 => 2,
+            _ => -1
+        };
+        if (!AdvanceSequence(order, caster.Position, WorldState.FutureTime(2f)))
+            ReportError($"unexpected order {order}");
+    }
+}
+
+class AhkMornInitial(BossModule module) : Components.StackWithIcon(module, (uint)IconID.AhkMornStack, ActionID.MakeSpell(AID.AkhMornCast), 6, 0, 2, 4);
```
xanunderscore, line 77: i think this whole component can be removed, right? since the AhkMornFollow component also watches for the icon addition. right now we get two "Stack!" hints in the vbm window

xanunderscore:
i just tested this out unsynced and it looks like the phantom kin's castbars aren't long enough to give AI mode time to dodge if you only use the cast for the telegraph, i got hit twice. might be worth showing the telegraph early, which is what BMR does - you can prob just copy it from there

LeontopodiumNivale14:
Alright, all the bugs should be fixed. Things tweaked here and there, names actually proper instead of 3am me coding and just... not double checking my work there.

PR #459, opened 2024-08-31
M3S Chain Deathmatch Danger Zones

sourpuh:
The green and yellow in M3S chain deathmatch blur together and it's difficult to discern what is actually safe. This change shows the inverse of the safe zone as a danger zone. This makes it significantly clearer where to stand for chain deathmatch 2.
cdmdangerzones

This PR is more of a suggestion than a perfect implementation. I couldn't figure out a good way to "inverse" the lariat AOE, so I sort of hacked it by flipping the origin and rotation and making it narrower. Also I didn't adjust the "Go into cleave!" hint which isn't displayed anymore with this change.


awgil:
That's a great idea! Thanks

PR #462, opened 2024-08-31
PotentialTargets enhancements

xanunderscore:
- add ActorType.Part, which is generally a part of a boss (tioman's wings, titan's heart, shinryu's tail)
- skip targets who are about to die (this is HUGE qol on casters or any class that has actions with very delayed damage application)
- set current-fate priority to 0 instead of -1

### DIFF
--- a/BossMod/BossModule/AIHints.cs
+++ b/BossMod/BossModule/AIHints.cs
@@ -90,17 +90,17 @@ public sealed class AIHints
     {
         bool playerInFate = ws.Client.ActiveFate.ID != 0 && ws.Party.Player()?.Level <= Service.LuminaRow<Lumina.Excel.GeneratedSheets.Fate>(ws.Client.ActiveFate.ID)?.ClassJobLevelMax;
         var allowedFateID = playerInFate ? ws.Client.ActiveFate.ID : 0;
-        foreach (var actor in ws.Actors.Where(a => a.Type is ActorType.Enemy or ActorType.Part && a.IsTargetable && !a.IsAlly && !a.IsDead))
+        foreach (var actor in ws.Actors.Where(a => a.IsTargetable && !a.IsAlly && !a.IsDead))
         {
             // fate mob in fate we are NOT a part of, skip entirely. it's okay to "attack" these (i.e., they won't be added as forbidden targets) because we can't even hit them
             // (though aggro'd mobs will continue attacking us after we unsync, but who really cares)
             if (actor.FateID > 0 && actor.FateID != allowedFateID)
                 continue;
-            
+
             // target is dying; skip it so that AI retargets, but ensure that it's not marked as a forbidden target
-            // skip this check on striking dummies as they die constantly
+            // skip this check on striking dummies (name ID 541) as they die constantly
             var predictedHP = ws.PendingEffects.PendingHPDifference(actor.InstanceID);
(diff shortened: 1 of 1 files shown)


BossMod/BossModule/AIHints.cs
```diff
@@ -90,16 +90,24 @@ public void FillPotentialTargets(WorldState ws, bool playerIsDefaultTank)
     {
         bool playerInFate = ws.Client.ActiveFate.ID != 0 && ws.Party.Player()?.Level <= Service.LuminaRow<Lumina.Excel.GeneratedSheets.Fate>(ws.Client.ActiveFate.ID)?.ClassJobLevelMax;
         var allowedFateID = playerInFate ? ws.Client.ActiveFate.ID : 0;
-        foreach (var actor in ws.Actors.Where(a => a.Type == ActorType.Enemy && a.IsTargetable && !a.IsAlly && !a.IsDead))
+        foreach (var actor in ws.Actors.Where(a => a.Type is ActorType.Enemy or ActorType.Part && a.IsTargetable && !a.IsAlly && !a.IsDead))
```
awgil, line 93: Is the type check even needed, now that we have IsAlly flag?

BossMod/BossModule/AIHints.cs
```diff
@@ -90,16 +90,24 @@ public void FillPotentialTargets(WorldState ws, bool playerIsDefaultTank)
...
         {
             // fate mob in fate we are NOT a part of, skip entirely. it's okay to "attack" these (i.e., they won't be added as forbidden targets) because we can't even hit them
             // (though aggro'd mobs will continue attacking us after we unsync, but who really cares)
             if (actor.FateID > 0 && actor.FateID != allowedFateID)
                 continue;
+            
+            // target is dying; skip it so that AI retargets, but ensure that it's not marked as a forbidden target
+            // skip this check on striking dummies as they die constantly
+            var predictedHP = ws.PendingEffects.PendingHPDifference(actor.InstanceID);
+            if (actor.HPMP.CurHP + predictedHP <= 0 && actor.OID != 0x23B)
```
awgil, line 103: Aren't there a whole bunch of different striking dummies with different id's?

BossMod/BossModule/AIHints.cs
xanunderscore: Oh maybe not. Are eventobjects allies?

BossMod/BossModule/AIHints.cs
xanunderscore: yeah good point, we could check for nameID == 541

BossMod/BossModule/AIHints.cs
awgil: I think so. Here's the check:
```
        var friendly = chr == null || ActionManager.ClassifyTarget(chr) != ActionManager.TargetCategory.Enemy;
```

BossMod/BossModule/AIHints.cs
awgil: Well do they all have the same nameid? :D

xanunderscore: Sheets only has "striking dummy" (541) and "timeworn striking dummy" (13078) (no idea what this is even)

the dummy at my house, the one in explorer mode, and the ones at Falcon's Nest in coerthas are all 541, so i'm gonna assume it applies to all

PR #469, opened 2024-09-02
Set auto record default on

xanunderscore:


BossMod/Replay/ReplayManagementConfig.cs
```diff
@@ -7,7 +7,7 @@ public class ReplayManagementConfig : ConfigNode
     public bool ShowUI = false;
 
     [PropertyDisplay("Auto record replays on duty start/end")]
-    public bool AutoRecord = false;
+    public bool AutoRecord = true;
```
awgil, line 10: So, the reason it was not auto enabled is that it will cause endless storage growth for unsuspecting people. Now, max-replays-to-keep can help with that, but what is the appropriate value?

xanunderscore: maybe 100? average replay size on my machine is about 500kb which would work out to 50MB total, the average ff14 player's mare directory is like 20 times that size.

PR #480, opened 2024-09-08
Add optional role-based colors to party members, and focus target

SomeRandomPeople0:
Pretty small change which might help with visual clarity for some people.
- Adds an option to color party members based on their role color, as long as their color would have been `ArenaColor.PlayerGeneric`
- Adds an option to specifically show the party member that is your focus target

I just kinda hard-coded the colors here, it might be better to make them configurable.


BossMod/BossModule/BossModule.cs
```diff
@@ -306,7 +306,15 @@ private void DrawPartyMembers(int pcSlot, Actor pc)
         foreach (var (slot, player) in Raid.WithSlot().Exclude(pcSlot))
         {
             var (prio, color) = CalculateHighestPriority(pcSlot, pc, slot, player);
-            if (prio == BossComponent.PlayerPriority.Irrelevant && !WindowConfig.ShowIrrelevantPlayers)
+
+            bool isFocus = false;
+            if (Service.TargetManager.FocusTarget != null)
```
awgil, line 311: This will break UIDev, and in general replays. Most of the classes in vbm should not access game state directly.

SomeRandomPeople0: I'm not super familiar with this codebase, would it be more appropriate change the behavior so that it checks if a party member is focus-targeted in `BossMod/Framework/WorldStateGameSync.cs` around `UpdateParty`? Something like storing the slot of the focus'd player?

awgil: Yeah, that's probably the right way - you'll need to add the focustarget to the world state (more precisely, i guess clientstate, since it's really player-specific) and check that here

awgil: I've added focus target to client state now.

SomeRandomPeople0: Thanks for getting to this, I was a bit busy the past couple days. I went ahead and adjusted it to use `WorldState.Client.FocusTargetId`. Let me know if I did this incorrectly.

BossMod/BossModule/BossModule.cs
```diff
@@ -318,6 +326,33 @@ private void DrawPartyMembers(int pcSlot, Actor pc)
...
+                    {
+                        color = ImGui.ColorConvertFloat4ToU32(new Vector4(1, 1, 0, 1));
+                    }
+                    else if (WindowConfig.ColorPlayersBasedOnRole)
+                    {
+                        switch (player.Role)
+                        {
+                            case Role.Melee:
+                            case Role.Ranged:
+                                color = ImGui.ColorConvertFloat4ToU32(new Vector4(1, 0, 0, 1));
```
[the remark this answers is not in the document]
SomeRandomPeople0: I adjusted the colors to be more muted, and put them in the ColorConfigs. I used `Service.Config.Get<ColorConfig>()` to get the instance, but I see other parts of the code use Timeline. Is this fine?

Aimsucks:
Updated my PR to include the changes @SomeRandomPeople0 made.

image
image

awgil:
Thanks folks, will take it from here!

PR #481, opened 2024-09-09
ops for pet state, duty action charges, blue mage spells

xanunderscore:
- pet state needed for scholar, used to check whether a Place order is currently active so we don't spam it trying to place Eos at fight start
- duty action charges are dynamically updated during some duties, specifically in this case the thancred vs ran'jit fight. the duty action should always be used ASAP if it has a charge, but if it doesn't have a charge, then pushing it onto the queue just locks up the rotation.
- BLU spells for obvious reasons

### DIFF
--- a/BossMod/Data/ClientState.cs
+++ b/BossMod/Data/ClientState.cs
@@ -1,6 +1,4 @@
-﻿using System.Windows.Markup;
-
-namespace BossMod;
+﻿namespace BossMod;
 
 public record struct ClientActionRequest
 (
@@ -39,6 +37,8 @@ public sealed class ClientState
     public record struct Gauge(ulong Low, ulong High);
     public record struct Stats(int SkillSpeed, int SpellSpeed, int Haste);
 
+    public record struct Pet(uint InstanceID, byte Order, byte Stance);
+
     public const int NumCooldownGroups = 82;
     public const int NumClassLevels = 32; // see ClassJob.ExpArrayIndex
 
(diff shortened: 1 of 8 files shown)


BossMod/Data/ClientState.cs
```diff
@@ -50,16 +50,28 @@ public record struct Stats(int SkillSpeed, int SpellSpeed, int Haste);
...
     public Fate ActiveFate;
+    public Pet ActivePet;
 
     public int ClassJobLevel(Class c)
     {
         var index = Service.LuminaRow<Lumina.Excel.GeneratedSheets.ClassJob>((uint)c)?.ExpArrayIndex ?? -1;
         return index >= 0 && index < ClassJobLevels.Length ? ClassJobLevels[index] : -1;
     }
 
+    public unsafe T GetGauge<T>() where T : unmanaged
```
awgil, line 66: Why is it needed here? This is a very ugly hack...

BossMod/Framework/ActionManagerEx.cs
```diff
@@ -167,12 +167,18 @@ public float GCD()
...
     {
         var id = ActionManager.GetDutyActionId(slot);
-        return id != 0 ? new(ActionType.Spell, id) : default;
+        if (id == 0)
+            return default;
+
+        var action = new ActionID(ActionType.Spell, id);
+        var dir = EventFramework.Instance()->GetContentDirector();
+
+        return new(action, dir == null ? default : *((byte*)dir + 1484 + slot));
```
awgil, line 179: This really needs to be added to CS, I'll forget about the existence of this field and stuff will break...

BossMod/Framework/WorldStateGameSync.cs
```diff
@@ -588,6 +588,18 @@ private unsafe void UpdateClient()
         var levels = uiState->PlayerState.ClassJobLevels;
         if (!MemoryExtensions.SequenceEqual(_ws.Client.ClassJobLevels.AsSpan(), levels))
             _ws.Execute(new ClientState.OpClassJobLevelsChange(levels.ToArray()));
+
+        Span<uint> blueSpells = stackalloc uint[24];
+        blueSpells.Clear();
+        for (var i = 0; i < 24; i++)
+            blueSpells[i] = actionManager->GetActiveBlueMageActionInSlot(i);
```
awgil, line 595: Can just compare actionmanager's BlueMageActions array directly, no need to call some function 24 times

BossMod/Data/ClientState.cs
```diff
@@ -50,16 +50,28 @@ public record struct Stats(int SkillSpeed, int SpellSpeed, int Haste);
     public Stats PlayerStats;
     public readonly Cooldown[] Cooldowns = new Cooldown[NumCooldownGroups];
     public readonly ActionID[] DutyActions = new ActionID[2];
+    public readonly byte[] DutyActionCharges = new byte[2];
```
awgil, line 53: I think this should actually be a single structure (with 2 action id's and 2 charge counts) rather than two different two-element arrays...

awgil:
Will merge tomorrow I guess, just need to fix these things...

BossMod/Data/ClientState.cs
xanunderscore: i put this in sch rotation:

```c#
        d.Spell(AID.EnergyDrain)!.ForbidExecute =
            (ws, _, _, _) => ws.Client.GetGauge<ScholarGauge>().Aetherflow == 0;
```

i could move it somewhere else i guess, but it's still going to involve an unsafe cast

BossMod/Data/ClientState.cs
xanunderscore: agreed, i didn't know we had versioned ops. i can change that

BossMod/Data/ClientState.cs
awgil: I see, ok, yeah it makes sense...

PR #487, opened 2024-09-15
ihuykatumu boss 1: widen persistent cone aoe

xanunderscore:
i had this set as 64 degrees in my local copy because i kept getting clipped by 60 degrees, even though that's the width of the visual. i think this might be slightly too wide but if it keeps people from getting hit then it's fine


awgil:
I think this is no longer needed - the issue was that actual cast origin for successive spells is slightly farther away than the boss, making aoe cover a bit more ground. This is now fixed.

PR #489, opened 2024-09-16
Dawntrail solo duties "A Father First" and "Feat of the Brotherhood"

xanunderscore:
also includes AI hinting for tower components.

i wrote out all the phase transitions in A Father First because i wanted to test picto downtime preparation, other than that it's extremely overkill

### DIFF
--- a/BossMod/Components/Towers.cs
+++ b/BossMod/Components/Towers.cs
@@ -2,13 +2,14 @@
 
 public class GenericTowers(BossModule module, ActionID aid = default) : CastCounter(module, aid)
 {
-    public struct Tower(WPos position, float radius, int minSoakers = 1, int maxSoakers = 1, BitMask forbiddenSoakers = default)
+    public struct Tower(WPos position, float radius, int minSoakers = 1, int maxSoakers = 1, BitMask forbiddenSoakers = default, DateTime activation = default)
     {
         public WPos Position = position;
         public float Radius = radius;
         public int MinSoakers = minSoakers;
         public int MaxSoakers = maxSoakers;
         public BitMask ForbiddenSoakers = forbiddenSoakers;
+        public DateTime Activation = activation;
 
         public readonly bool IsInside(WPos pos) => pos.InCircle(Position, Radius);
         public readonly bool IsInside(Actor actor) => IsInside(actor.Position);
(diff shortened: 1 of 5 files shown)


BossMod/Util/ShapeDistance.cs
```diff
@@ -199,6 +199,6 @@ Func<WPos, float> edge((WPos p1, WPos p2) e)
     }
     public static Func<WPos, float> ConvexPolygon(IEnumerable<WPos> vertices, bool cw, float offset = 0) => ConvexPolygon(PolygonUtil.EnumerateEdges(vertices), cw, offset);
 
-    public static Func<WPos, float> Intersection(List<Func<WPos, float>> funcs, float offset = 0) => p => funcs.Max(e => e(p)) - offset;
-    public static Func<WPos, float> Union(List<Func<WPos, float>> funcs, float offset = 0) => p => funcs.Min(e => e(p)) - offset;
+    public static Func<WPos, float> Intersection(IEnumerable<Func<WPos, float>> funcs, float offset = 0) => p => funcs.Max(e => e(p)) - offset;
+    public static Func<WPos, float> Union(IEnumerable<Func<WPos, float>> funcs, float offset = 0) => p => funcs.Min(e => e(p)) - offset;
```
awgil, line 203: This function is called multiple (thousands!) of times, so the reason it was a list is for performance (we don't want to enumerate arbitrarily complex enumerable multiple times, plus technically enumerable could even be single-pass).

BossMod/Components/Towers.cs
```diff
@@ -46,6 +47,34 @@ public override void AddHints(int slot, Actor actor, TextHints hints)
...
+    public override void AddAIHints(int slot, Actor actor, PartyRolesConfig.Assignment assignment, AIHints hints)
+    {
+        // draw hints for imminent, currently unsoaked towers
+        // we consider some list of towers part of the same "group" if their activations are within 500ms, as there can be varying delays between helper actors in an encounter casting the "same" spell
+        // generally, successive towers that are meant to be soaked by one player (i.e. in quest battles) activate no more frequently than about 2 seconds apart, so this is pretty conservative
+        var imminentMissedTowers = Towers.Where(t => ShouldHint(slot, actor, t))
+            .GroupBy(x => x.Activation.Round(TimeSpan.FromMilliseconds(500)))
+            .MinBy(gt => gt.Key);
+
+        if (imminentMissedTowers != null)
```
awgil, line 59: If I understand correctly, it could just be done in a way similar to other components:
```
var deadline = Towers[0].Activation.AddSeconds(0.5);
List<Func<WPos, float>> imminentTowers = [.. Towers.TakeWhile(t => t.Activation <= deadline).Select(t => ShapeDistance.InvertedCircle(t.Position, t.Radius)];
hints.AddForbiddenZone(ShapeDistance.Intersection(imminentTowers);
```

PR #501, opened 2024-09-25
Ports

Herculezz55:
removed DohnMehg


BossMod/Components/InterceptTether.cs
```diff
@@ -0,0 +1,62 @@
+﻿namespace BossMod.Components;
+
+// generic component for tethers that need to be intercepted eg. to prevent a boss from gaining buffs - (Ported From BMR)
+public class InterceptTether(BossModule module, ActionID aid, uint tetherID) : CastCounter(module, aid)
```
awgil, line 4: How is it different from existing tether components (tankbustertether / baitawaytether)? No aoe, can target both other enemies and players, anything else?

BossMod/Components/LineOfSightAOE.cs
```diff
@@ -9,6 +11,9 @@ public abstract class GenericLineOfSightAOE(BossModule module, ActionID aid, flo
     public WPos? Origin { get; private set; } // inactive if null
     public List<(WPos Center, float Radius)> Blockers { get; private set; } = [];
     public List<(float Distance, Angle Dir, Angle HalfWidth)> Visibility { get; private set; } = [];
+    public List<AOEInstance> Safezones = [];
+    public List<AOEShape> UnionShapes = [];
+    public List<AOEShape> DifferenceShapes = [];
```
awgil, line 16: Why do we need all this?

BossMod/Components/InterceptTether.cs
CarnifexOptimus: It is for tethers that simply need to be intercepted to prevent something (eg. boss gaining buffs)

BossMod/Components/LineOfSightAOE.cs
CarnifexOptimus: he is missing all the custom AOEShape logic, it wont work like this. i basically refactored your whole line of sight component to be more accurate, support rectangle aoes etc.

https://github.com/FFXIV-CombatReborn/BossmodReborn/blob/main/BossMod/Components/LineOfSightAOE.cs

CarnifexOptimus:
I also doubt my dohn mheg boss 3 module will work properly without porting the custom waypoints over somehow... how is the AI supposed to navigate the chasm without.

BossMod/Modules/Stormblood/Dungeon/D03BardamsMettle/D032HunterOfBardam.cs
```diff
@@ -0,0 +1,183 @@
...
+}
+
+public enum IconID : uint
+{
+    BardamsRing = 58, // player
+    ChasingAOE = 197, // player
+}
+
+class Comet(BossModule module) : Components.LocationTargetedAOEs(module, ActionID.MakeSpell(AID.CometFirst), 4);
+class CometFirst(BossModule module) : Components.LocationTargetedAOEs(module, ActionID.MakeSpell(AID.CometFirst), 4);
```
xanunderscore, line 54: why do we have two components for CometFirst?

xanunderscore:
the dohn mheg B3 module should probably be removed from the PR, we either need some tweaks for path rasterization in that arena OR the level collision as forbidden zones feature for it to work

BossMod/Modules/Stormblood/Dungeon/D03BardamsMettle/D032HunterOfBardam.cs
CarnifexOptimus: seems like he copied it incorrectly, on BMR it is 
class Comet(BossModule module) : Components.StandardChasingAOEs(module, new AOEShapeCircle(4), ActionID.MakeSpell(AID.CometFirst), ActionID.MakeSpell(AID.CometRest), 10, 1.5f, 9, true, (uint)IconID.ChasingAOE);
class CometFirst(BossModule module) : Components.LocationTargetedAOEs(module, ActionID.MakeSpell(AID.CometFirst), 4);
class CometRest(BossModule module) : Components.LocationTargetedAOEs(module, ActionID.MakeSpell(AID.CometRest), 4);

but on vbm the AI for chasingAOEs is broken anyway, it will often get stuck inside AOEs (this affects both old and new pathfinding algorithm) because it cant decide where to go. i made sort of a hack to circumvent the problem

    public override void AddAIHints(int slot, Actor actor, PartyRolesConfig.Assignment assignment, AIHints hints)
    {
        if (Chasers.Count > 0)
        {
            AddForbiddenZones(actor, hints, true);
            AddForbiddenZones(actor, hints, false);
        }
    }

    private void AddForbiddenZones(Actor actor, AIHints hints, bool isTarget)
    {
        // sort of a hack to prevent the AI from getting "stuck" inside the AOE because all paths to safety have equal distance
        foreach (var c in Chasers.Where(c => c.Target == actor == isTarget))
        {
            var circle = (AOEShapeCircle)c.Shape;
            var radius = isTarget ? MoveDistance + circle.Radius : circle.Radius + 1;
            var position = isTarget ? c.PredictedPosition() - circle.Radius * actor.Rotation.ToDirection() : c.PredictedPosition();
            hints.AddForbiddenZone(ShapeDistance.Circle(position, radius), c.NextActivation);
        }
    }

CarnifexOptimus:
> the dohn mheg B3 module should probably be removed from the PR, we either need some tweaks for path rasterization in that arena OR the level collision as forbidden zones feature for it to work

I don't think collision alone will help. On BMR i accurately rebuilt the chasm arena using collision data, it still falls off without managing the waypoints manually. both with new and old pathfinding algorithm. i think the rasterisation is just not accurate enough for such tiny chasms.

BossMod/Modules/Stormblood/Dungeon/D03BardamsMettle/D032HunterOfBardam.cs
xanunderscore: yeah my version of bardam's doesn't even use chasing AOEs, they have a long enough cast time that ai can always easily avoid them

Herculezz55: Yes I modified it last minute to remove the chasing aorncuae that shit didn't work and as you said there's plenty of time to avoid this way. Can r3move the duplicate.

Herculezz55:
> I also doubt my dohn mheg boss 3 module will work properly without porting the custom waypoints over somehow... how is the AI supposed to navigate the chasm without. You either need to port it into AIhint forcedmovement (i dont think it accepts a list at the moment, so would need to check if AI arrived at point and set the next waypoint). or port over the whole waypoint manager: https://github.com/FFXIV-CombatReborn/BossmodReborn/blob/main/BossMod/Pathfinding/WaypointManager.cs which requires hints.WaypointManager.UpdateCurrentWaypoint(player.Position);
> 
> ```
>     if (hints.WaypointManager.HasWaypoints)
>     {
>         var currentWaypoint = hints.WaypointManager.CurrentWaypoint;
>         if (currentWaypoint.HasValue)
>         {
>             return new NavigationDecision
>             {
>                 Destination = currentWaypoint.Value,
>                 LeewaySeconds = float.MaxValue,
>                 TimeToGoal = (currentWaypoint.Value - player.Position).Length() / playerSpeed,
>                 DecisionType = Decision.CustomWaypoints
>             };
>         }
>     }
> ```
> 
> to be added to the navigationdecision.cs and aihints.cs needs a public WaypointManager WaypointManager { get; private set; } = new WaypointManager();

Na I removed this because it wasn't working, and for AD purposes the support kills the lyre just fine so we can just wait till they are done, at least until someone figures out a batter way to do this.

CarnifexOptimus:
> > I also doubt my dohn mheg boss 3 module will work properly without porting the custom waypoints over somehow... how is the AI supposed to navigate the chasm without. You either need to port it into AIhint forcedmovement (i dont think it accepts a list at the moment, so would need to check if AI arrived at point and set the next waypoint). or port over the whole waypoint manager: https://github.com/FFXIV-CombatReborn/BossmodReborn/blob/main/BossMod/Pathfinding/WaypointManager.cs which requires hints.WaypointManager.UpdateCurrentWaypoint(player.Position);
> > ```
> >     if (hints.WaypointManager.HasWaypoints)
> >     {
> >         var currentWaypoint = hints.WaypointManager.CurrentWaypoint;
> >         if (currentWaypoint.HasValue)
> >         {
> >             return new NavigationDecision
> >             {
> >                 Destination = currentWaypoint.Value,
> >                 LeewaySeconds = float.MaxValue,
> >                 TimeToGoal = (currentWaypoint.Value - player.Position).Length() / playerSpeed,
> >                 DecisionType = Decision.CustomWaypoints
> >             };
> >         }
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to be added to the navigationdecision.cs and aihints.cs needs a public WaypointManager WaypointManager { get; private set; } = new WaypointManager();
> 
> Na I removed this because it wasn't working, and for AD purposes the support kills the lyre just fine so we can just wait till they are done, at least until someone figures out a batter way to do this.

might work for AD with NPCs, but not for multi boxing users
if it isn't working its probably missing something somewhere, because it works on BMR and there is no reason it wouldnt on vbm. is the error already happening while compiling?

BossMod/Components/InterceptTether.cs
Herculezz55: > How is it different from existing tether components (tankbustertether / baitawaytether)? No aoe, can target both other enemies and players, anything else?

I tried using tankbustertether for this mechanic but couldn't get it to work. I'm still a newbie with vbm ;P

BossMod/Modules/Stormblood/Dungeon/D03BardamsMettle/D032HunterOfBardam.cs
CarnifexOptimus: well BMR got many components modified and framework changes, so just copying the modules alone will not always give the intended results.
i know it can technically be dodged by the cast time alone, 1.5s is plenty for AI when the circle is only 5 diameter (needs about 833ms to run out). But i make my modules not just for a single use case in mind so I added it for non AI users since they dont have 1 frame reaction times like AI.

BossMod/Components/LineOfSightAOE.cs
Herculezz55: It works they way I ported it, my porting method is to donthe least framework modification ad possible.

BossMod/Components/InterceptTether.cs
Herculezz55: removed this module

PR #509, opened 2024-09-26
Funny bridge

xanunderscore:
i actually got toaded while running this so i need to figure out what's wrong with the cone

### DIFF
--- a/BossMod/Modules/Shadowbringers/Dungeon/D02DohnMheg/D023AencThon.cs
+++ b/BossMod/Modules/Shadowbringers/Dungeon/D02DohnMheg/D023AencThon.cs
@@ -51,7 +51,7 @@ public enum SID : uint
 class CripplingBlow(BossModule module) : Components.SingleTargetCast(module, ActionID.MakeSpell(AID.CripplingBlow));
 class VirtuosicCapriccio(BossModule module) : Components.RaidwideCast(module, ActionID.MakeSpell(AID.VirtuosicCapriccio));
 class ImpChoir(BossModule module) : Components.CastGaze(module, ActionID.MakeSpell(AID.ImpChoir));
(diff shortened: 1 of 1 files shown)


CarnifexOptimus:
i have a suspicion what could be wrong with the cone since it only happens with the new pathfinding. the new pathfinding only checks the 4 corners of a cell (correct me if i understand the code of it incorrectly), so something can poke in between the corners, for example the pointy part of a cone. this would result in the pixel incorrectly marked as safe.

xanunderscore:
no the problem with pathfinding for this boss was just that the g-score calculation for forbidden zones was not used for arena bounds, arena bounds just checked the center of each pixel. veyn fixed that so now pathfinding on the bridge (with sufficient map resolution) works fine

also you are correct, i completely forgot about the second half of the fight, i was mainly just putting this up in case veyn was interested in seeing why the pathfinding perf isn't great during bridge phase

CarnifexOptimus:
i still got hit by the cone yesterday when testing the waypoint manager in vbm code, the ai stood inside the boss hitbox and maybe slightly touched the pointy end of the cone and got hit.

the reason the performance is bad might be because you increased grid resolution to 0.25 if i am seeing this correctly. that makes pathfinding 4 times as expensive to run

xanunderscore:
i doubt it, 20x20 at 0.25 resolution is only 1.7 times the size of the default grid (30x30 at 0.5) and the impact of trash mob/fate pathfinding on my FPS is unnoticeable, i think it's more likely to do with the length of the path

CarnifexOptimus:
the total grid size doesn't matter much as far as I can tell. its the amount of cells it has to check until destination, which goes up quadratically when increasing the resolution. even a 100x100 grid will likely still have no noticeable performance impact after its initial creation, but mostly the distances between you and destination. this is also why extremly big AOEs, even if its only a rectangle, can tank the performance, since it has to check a lot of cells to find the optimal path out of the aoe.

PR #518, opened 2024-10-04
standardise the windows and commands

Jaksuhn:
removed the drawui configs since that was kind of a weird way of going about it

added close buttons to ai/ar and somewhat standardised their commands to be like the replay window

added buttons to the about tab to open those windows since those configs were the only non command way of doing it before

I had to change the access level of a few things in the main file to add those buttons. Would've had to pass a bunch of stuff around to configui then to abouttab otherwise. Not sure which you'd prefer

### DIFF
--- a/BossMod/AI/AIConfig.cs
+++ b/BossMod/AI/AIConfig.cs
@@ -9,9 +9,6 @@ sealed class AIConfig : ConfigNode
     [PropertyDisplay("Enable AI", tooltip: "Disclaimer: AI is very experimental, use at your own risk!")]
     public bool Enabled = false;
 
-    [PropertyDisplay("Show in-game UI")]
-    public bool DrawUI = true;
-
     [PropertyDisplay("Show advanced options in the UI")]
     public bool ShowExtraUIOptions = true;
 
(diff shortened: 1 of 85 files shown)


awgil:
Ok, so what's the problem with 'show UI' settings in the config? This gives persistence automatically - and with this approach you'd have to reopen (or re-close) each window every time you log in.

No worries about adding close buttons.

Jaksuhn:
> Ok, so what's the problem with 'show UI' settings in the config? This gives persistence automatically

The persistence is really the only pro imo. A lot of settings are just bloated and have double "enable this thing" configs in addition to the thing itself (see auto peloton) which is just confusing. I can at least revert that part. I assume the rest was ok?

awgil:
Well - if you move 'show ui' flags back to the config, then you won't need to make any static/internal/etc changes - the window's OnClose can simply toggle the 'show ui' to false (see ReplayManagementWindow for how it would work). I would first make this change, and then see whether it's still causing confusion - if so, will think of how to handle that.

Jaksuhn:
closing in favour of veyn's plans mentioned in the server

awgil:
Implemented in v255.

PR #522, opened 2024-10-06
fix file exists check in PresetDatabase

xanunderscore:
the `Exists` property is cached so this code would never work


awgil:
> the `Exists` property is cached so this code would never work

Uhhhh lol wtf... This is crazy

awgil:
Indeed...
> When first called, [FileInfo](https://learn.microsoft.com/en-us/dotnet/api/system.io.fileinfo?view=net-8.0) calls [Refresh](https://learn.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo.refresh?view=net-8.0) and caches information about the file. On subsequent calls, you must call [Refresh](https://learn.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo.refresh?view=net-8.0) to get the latest copy of the information.

PR #523, opened 2024-10-07
ChangelogNotice + remove AI options

xanunderscore:
a lot of these were not needed. also i made ChangelogNotice for displaying arbitrary info in the changelog popup seeing as you can't put a "since" attribute on an option that no longer exists.

<details>
<summary>Screenshots</summary>

</details>


BossMod/BossModule/ZoneModuleRegistry.cs
```diff
@@ -26,9 +26,9 @@ static ZoneModuleRegistry()
                 Service.Log($"Zone module {t} has no ZoneModuleInfo attribute, skipping");
                 continue;
             }
-            if (_modulesByCFC.ContainsKey(attr.CFCID))
```
xanunderscore, line 29: btw this change is only here because this file fails the code style check otherwise

PR #524, opened 2024-10-07
ensure existence of default preset

xanunderscore:
based on a discussion in discord, it seems like there can be a few factors that prevent people from even knowing that vbm includes a sane-ish default preset - one being the `Exists` bug, but also plenty of people who have been using the plugin for a long time have gotten grandfathered in to the older, shittier system where we had two separate default presets, or no default at all. and since PresetDatabase would only seed the default if the preset file was missing entirely (which, for any existing users, it wouldn't be), the vast majority of those users never found out that these defaults existed.

in this PR, i ensure that we always load the default presets, which i renamed to `VBM Default` to clarify what it is and because i already have my own `Default` that i'm using. i also added an option to hide it from the UI, since power users probably won't want it.

here's how it looks if you try to edit the default:

image

tooltip for the delete button: 
image

### DIFF
--- a/BossMod/Autorotation/AutorotationConfig.cs
+++ b/BossMod/Autorotation/AutorotationConfig.cs
@@ -19,6 +19,9 @@ public sealed class AutorotationConfig : ConfigNode
     [PropertyDisplay("Show autorotation preset in the server info bar")]
     public DtrStatus ShowDTR = DtrStatus.None;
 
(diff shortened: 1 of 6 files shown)


awgil:
Good stuff, i'll probably replace name-based criterion with some explicit 'is system' flag on a preset, as I don't want to think of what would happen if user would call his custom preset "VBM Default" for shits and giggles.

PR #533, opened 2024-10-19
Treasure Dungeon Maturity + POTD Modules

LeontopodiumNivale14:
Accidentally left the maturity of the module as a WIP, when it was very much done. Noticed it while I was doing treasure dungeons just now XD


LeontopodiumNivale14:
To better summarize all of this at once (because it's a lot)
-> Originally updated the maturity of the floor 100 treasure dungeon boss 1 (still missing boss 2)

POTD wise:
-> Created floors 160, 170, 180, 190.
-> Updated floor 80 to match the standard that I created on floor 180, and added a hint system that tells you what the following attacks are going to be/where the cleaves that behemoth does on both floor sets (both have a different rotation of attacks they do before they do the cleave, and floor 80 cleaves E -> W -> E.. vs floor 180 does S -> N -> S...
-> Floor 190 wise added an enrage timer for the bombs that need to be killed, and also while the boss is casting the attack that needs stunned. 
-> Last thing was added the list of waypoints that they stun bomb spawns on floor 190, need to go back through when I get some free time and add a minimap hint where the stun bomb will be, that way if desired can pre-position the boss onto it to not have to worry about trying to KB the ad into the boss.

PR #542, opened 2024-11-06
Antitower, Sohr Khai, Xelphatol

xanunderscore:
notable framework changes:

1. i made the eye drawing logic a public static method of GenericGaze. this is because Calcabrina (Antitower B3) has a gaze attack with an associated shape, namely a cone. i figured it would be less risky to roll my own "conditional" gaze component than it would be to refactor the existing gaze component to support a shape, but i'm open to making that change instead if you think it would be better
2. added AffectedPlayers property to LOS component, so i could use it for Antitower B2's tether mechanic where one player needs to hide behind a meteor.

also we should consider using castinfo location instead of caster location for self targeted AOEs. encounters where I noticed castinfo location differing drastically from caster location are Sohr Khai B3, Hashmal (from the ivalice questline and one of the zadnor CEs), double dragons in DSR, and Lunar Odin in Death Unto Dawn

### DIFF
--- a/BossMod/Components/Gaze.cs
+++ b/BossMod/Components/Gaze.cs
@@ -48,11 +48,7 @@ public abstract class GenericGaze(BossModule module, ActionID aid = new(), bool 
         {
             bool danger = HitByEye(pc, eye) != Inverted;
             var eyeCenter = IndicatorScreenPos(eye.Position);
-            var dl = ImGui.GetWindowDrawList();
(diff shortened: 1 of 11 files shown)


awgil:
> also we should consider using castinfo location instead of caster location for self targeted AOEs

This would be a breaking change with unknown implications. I think way forward is to make a new better 'simple aoe' component that can replace both self-targeted and location targeted (use castinfo location for both) and deprecate old ones. Then existing components can be modified over time to use new component instead.

PR #549, opened 2024-11-25
Leylines module

Taurenkey:
Adds a module for staying within Leylines.

@awgil Not sure if you'd prefer this as a Misc AI or if it should be moved into BLM Utility? Lemme know.


awgil:
> Not sure if you'd prefer this as a Misc AI or if it should be moved into BLM Utility? Lemme know.

That's a good question, I don't know, since I don't know anything about BLM. For now, misc ai sounds reasonable, since it doesn't look like a pure utility? Or would you realistically put that into a cd planner?

PR #560, opened 2024-12-30
add thordan ex

Jaksuhn:
so since extremes and unreals are exactly the same, with just an enum difference, idk if you'd rather have a different approach than copy-pasting from one to the other and duplicating all the classes

also there's an issue with this/the unreal too where if you cause the first phase to end quick enough, intermission won't trigger


Jaksuhn:
actually on second thought a lot is broken by phase skipping since this doesn't have proper phases, this'll need to be changed drastically

PR #566, opened 2025-01-10
Auto autos: prevent user from manually restarting autos with RMB

xanunderscore:
See title. this got me killed in Eureka Orthos


xanunderscore:
nvm there's a better sig for this

xanunderscore:
there was not, in fact, a better sig

PR #572, opened 2025-01-21
Add DD to WorldState, pomanders and magicite to ActionID

xanunderscore:
I didn't add poms/magicite to actionqueue yet but it doesn't really matter since no modules will be using them. I had a concern about HandleActionRequest but it looks like we're ok with just directly calling it after we call the original function, same as UseBozjaFromHolsterDirectorDetour

### DIFF
--- a/BossMod/Data/DeepDungeonState.cs
+++ b/BossMod/Data/DeepDungeonState.cs
@@ -1,0 +1,180 @@
+﻿using static FFXIVClientStructs.FFXIV.Client.Game.InstanceContent.InstanceContentDeepDungeon;
+
+namespace BossMod;
+
+public sealed class DeepDungeonState
+{
+    public DungeonProgress Progress;
+    public byte DungeonId;
+    public RoomFlags[] MapData = new RoomFlags[25];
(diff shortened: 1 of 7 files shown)


BossMod/Data/DeepDungeonState.cs
```diff
@@ -0,0 +1,180 @@
...
+    public record struct Item(byte Count, byte Flags)
+    {
+        public readonly bool Usable => (Flags & (1 << 0)) != 0;
+        public readonly bool Active => (Flags & (1 << 1)) != 0;
+    }
+    public record struct Chest(byte Type, byte Room);
+
+    public Item GetItem(PomanderID pid) => GetSlotForPomander(pid) is var s && s >= 0 ? Items[s] : default;
+
+    public int GetSlotForPomander(PomanderID pid) => Service.LuminaRow<Lumina.Excel.Sheets.DeepDungeon>(DungeonId)!.Value.PomanderSlot.ToList().FindIndex(p => p.RowId == (uint)pid);
```
awgil, line 39: This looks inefficient, why do you convert to list here?

PR #578, opened 2025-01-27
some Utility fixes

Akechi-kun:
man I wrote some jank ass shit a few months back, I should revisit these files soon


BossMod/Autorotation/Utility/ClassBRDUtility.cs
```diff
@@ -29,11 +29,12 @@ public static RotationModuleDefinition Definition()
     public override void Execute(StrategyValues strategy, ref Actor? primaryTarget, float estimatedAnimLockDelay, bool isMoving)
     {
         ExecuteShared(strategy, IDLimitBreak3, primaryTarget);
-        ExecuteSimple(strategy.Option(Track.WardensPaean), BRD.AID.WardensPaean, Player);
+        ExecuteSimple(strategy.Option(Track.WardensPaean), BRD.AID.WardensPaean, ResolveTargetOverride(strategy.Option(Track.WardensPaean).Value) ?? primaryTarget ?? Player);
         ExecuteSimple(strategy.Option(Track.NaturesMinne), BRD.AID.NaturesMinne, Player);
 
         var troub = strategy.Option(Track.Troubadour);
-        if (troub.As<TroubOption>() != TroubOption.None)
+        var hasDefensive = StatusDetails(Player, BRD.SID.Troubadour, Player.InstanceID).Left > 5f || StatusDetails(Player, DNC.SID.ShieldSamba, Player.InstanceID).Left > 5f || StatusDetails(Player, MCH.SID.Tactician, Player.InstanceID).Left > 5f;
```
awgil, line 36: Eh, I'm not sure about this change:
1. First of all, you're checking a buff that has source-id equal to player - since troubadour has single charge, it's not possible for it to return true (and also definitely not possible to have dnc/mch buffs from brd player). I imagine you wanted to check any defensive here.
2. 5s limit is very arbitrary. I understand the intention, but conceivably you might want to override the buff anyway (eg there are situations where two raidwides can be covered by a single defensive if it's pressed late enough, and if some other guy pressed his defensive too early, arguably it's better to override the buff rather than not covering the second raidwide with a save.

I suggest adding another option to phys-ranged buffs 'use unless equivalent buff is up'.

Issue #543, opened 2024-11-15
Bug: Manual use of Bozja holster items does not go through MAO


xanunderscore:
as discussed on discord, this function https://github.com/awgil/ffxiv_bossmod/blob/master/BossMod/Framework/ActionManagerEx.cs#L451 doesn't call MAO at all, it calls the original director function directly. this results in, e.g., not being able to manually use light curtain if autorotation is on

xanunderscore:
While debugging this I also discovered that `ActionManager->IsRecastTimerActive` returns the wrong value if you pass it a bozja holster item, which makes sense, but this means that queueing Light Curtain will result in a lot of log spam since AMEx doesn't know not to try to use it before the GCD is up

PR #580, opened 2025-01-29
WIP deep dungeon auto clear modules

xanunderscore:
note that, because PalacePal doesn't have an IPC and it sounded like a huge PITA to add one, and it wouldn't even work in UIDev, the new zone module's trap avoidance functionality relies on just directly opening the PP sqlite file and reading trap locations from there.

i'm not a big fan of this solution but also like, i have no idea how to do it better, other than copying PP's entire dataset into the vbm source tree and assuming it will never change

In addition to some changes to the boss modules themselves since they mostly suck:

### UIDev
- draws in a separate movable window now
- added support to read FFXIV game path from environment variable - mine is installed into a custom location via steam and it doesn't seem like steam exposes any means of reporting custom paths for games
- added shitty "offline texture provider" to be able to draw the DD minimap in the replay viewer in UIDev. the red and blue channels are swapped because of an endianness mismatch or something. i didn't really feel like putting in the extra effort of converting them but i could if someone really wants it

### framework
- added replay "memory" support to replay viewer, it will reopen the previous replay to the same playback position if desired
- added Source attribute to ForbiddenZones, which for now is just the caster's instanceID, if one exists
- ActionManagerEx now checks for action LOS before blocking movement in preparation for a cast
- added InstanceContentDeepDungeon pomander and magicite use functions to CS, we can hook these and add to manual queue at some point if needed
- added code that fixes multiple charge GCDs that are conditionally affected by haste, of which only one exists, which is Drill
- improved interact-with-target code
  - added a slightly larger zone around target object to fix an edge case in the thancred solo duty in EW
  - switched to the interact function in TargetSystem that doesn't actually change your hard target as that's unnecessary
  - clear forcedmovement during the actual interact window to prevent immediate interruption of the interact action

### ui
- added separate Zone Module Window which can optionally be drawn. this is a regular imgui window. vbm's minimap is great, but i have it configured to ignore mouse input, which may not be desirable for zone modules. if present, the zone module window will also display in the replay viewer
- if the Zone Module Window wants to be drawn and there is no active boss module, the hints window will not appear (since its contents are expected to be drawn in the ZM window instead)

### other
fixed/improved some of my rotation code
- dumb oversight in PLD caused it to sometimes never use divine might holy spirit even in a buff window
- reworked DRG, MCH, and MNK to use `NextGCD`, a previously wasted feature, to correctly predict reassemble/life surge/true north usage
- took out the cast time stuff, replaced with the new framework-level solution


.github/workflows/build.yml
```diff
@@ -0,0 +1,44 @@
+name: Build
```
awgil, line 1: Can you explain what's this? What's `wip` branch and why do you want to run CI on every push?

BossMod/ActionQueue/ActionDefinition.cs
```diff
@@ -120,7 +120,13 @@ public float MainReadyIn(ReadOnlySpan<Cooldown> cooldowns, ReadOnlySpan<ClientSt
...
             return 0;
         var cdg = cooldowns[ActualMainCooldownGroup(dutyActions)];
-        return !IsMultiCharge || cdg.Total < Cooldown ? cdg.Remaining : Cooldown - cdg.Elapsed;
+        var max = MaxChargesAtCap();
+
+        // GCDs with multiple charges can be affected OR unaffected by haste depending on how many charges the user currently has access to
+        // the only separate-cooldown GCD that increases to >1 charge via trait is currently MCH Drill; others have multiple charges at unlock - SGE Phlegma, RPR Soul Slice, BLU Surpanakha
+        var cooldownSingleCharge = IsGCD && max > 1 && cdg.Total > 0 ? cdg.Total / max : Cooldown;
+
+        return !IsMultiCharge || cdg.Total < cooldownSingleCharge ? cdg.Remaining : cooldownSingleCharge - cdg.Elapsed;
```
awgil, line 129: Ok, I don't understand this change... Can you elaborate?

BossMod/Autorotation/MiscAI/AutoEngage.cs
```diff
@@ -0,0 +1,38 @@
+using BossMod.Autorotation.xan;
```
awgil, line 1: That's the same thing as in other PR?

BossMod/BossMod.csproj
```diff
@@ -18,6 +18,7 @@
     <ProduceReferenceAssembly>false</ProduceReferenceAssembly>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
awgil, line 21: What's this?

BossMod/BossModule/ZoneModule.cs
```diff
@@ -19,6 +19,8 @@ protected virtual void Dispose(bool disposing)
 
     public virtual void Update() { }
     public virtual bool WantToBeDrawn() => false; // return true if it wants to be drawn (higher priority than inactive boss modules, but lower priority than active)
+    public virtual bool DrawSeparately => false; // return true if the contents of DrawExtra should be shown in a separate window
```
awgil, line 22: Do we really need two modes? What are the usecases for both options?

BossMod/Data/Actor.cs
```diff
@@ -134,6 +135,8 @@ public sealed class Actor(ulong instanceID, uint oid, int spawnIndex, string nam
     public bool PredictedDead => PredictedHPRaw <= 1 && !IsStrikingDummy;
     public float PredictedHPRatio => (float)PredictedHPRaw / HPMP.MaxHP;
 
+    public bool IsTransformed => Statuses.Any(Autorotation.RotationModuleManager.IsTransformStatus);
```
awgil, line 138: I think these statuses need to be moved to `Data\` (something like CommonStatusId or something...). This is a logical dependency inversion here.

BossMod/ActionQueue/ActionDefinition.cs
xanunderscore: If you have the level 94 trait Enhanced Multiweapon on machinist, Drill stops being affected by skill speed/haste/slows, so the separate cooldown is always 20s and using both charges will set it to 40. before then it is affected by haste despite still technically having multiple charges, so like in Palace of the Dead, where you have some inherent skillspeed from the aetherpool gear, the single charge cooldown of Drill is 19.2s and using the one charge you have at that level will set the cooldown to 38.4s. also, if you use a single charge of drill at these levels while slowed or hasted, you can get a cooldown total of like 32 or 54 seconds, and checking Cooldown against that gives you incorrect results.

so what this line is doing is just dividing cdg.Total by the maximum number of charges of the ability, which SHOULD give us the effective single charge CD

BossMod/BossMod.csproj
xanunderscore: Without this property set, the plugin will fail to load because it won't be able to find the sqlite3 DLL

BossMod/Framework/ActionManagerEx.cs
```diff
@@ -481,6 +513,17 @@ private bool UseBozjaFromHolsterDirectorDetour(PublicContentBozja* self, uint ho
         return res;
     }
 
+    // TODO add to manual q
+    private void UsePomanderDetour(InstanceContentDeepDungeon* self, uint slot)
```
awgil, line 517: Oh, I remember you also mentioned some bozja thing not getting hooked?..

BossMod/BossModule/ZoneModule.cs
xanunderscore: The only usecase in master atm is the Sastasha module, which draws one line of hints in the regular bossmod hints window, and I felt like drawing a whole separate window for that would probably be overkill

BossMod/Framework/ActionManagerEx.cs
xanunderscore: https://github.com/awgil/ffxiv_bossmod/issues/543

BossMod/Replay/ReplayManager.cs
```diff
@@ -69,12 +73,22 @@ public void Show()
...
+    private readonly RotationDatabase rotationDB;
     private int _nextAnalysisId;
     private string _path = "";
     private FileDialog? _fileDialog;
+    private string fileDialogStartPath;
+
+    public ReplayManager(RotationDatabase rotationDB, string fileDialogStartPath)
+    {
+        this.rotationDB = rotationDB;
+        this.fileDialogStartPath = fileDialogStartPath;
```
awgil, line 85: Naming convention (_ prefix for private variables), no `this.` please, it's too ugly.

BossMod/Replay/ReplayManagementConfig.cs
```diff
@@ -27,4 +29,12 @@ public class ReplayManagementConfig : ConfigNode
...
     [PropertyDisplay("Format for recorded logs")]
     public ReplayLogFormat WorldLogFormat = ReplayLogFormat.BinaryCompressed;
+
+    [PropertyDisplay("Open previously open replays on plugin reload")]
+    public bool RememberReplays;
+
+    [PropertyDisplay("Remember playback position for previously opened replays")]
+    public bool RememberReplayTimes;
+
+    public List<ReplayMemory> ReplayHistory = [];
```
awgil, line 39: Not a fan of storing this 'invisible' state in configs tbh. At very least, it would mean that config would always be 'modified' in UIDev, preventing close by esc.

I wonder if we need a separate config (user preferences or something), it could also be used to store shit like default filtered statuses in planner, default filtered out actions for replays, etc etc etc.

BossMod/Util/WPosDir.cs
```diff
@@ -53,6 +53,7 @@ public readonly bool InRect(WDir direction, float lenFront, float lenBack, float
 }
 
 // 2d vector that represents world-space position on XZ plane
+[Serializable]
```
awgil, line 56: Huh?

BossMod/BossMod.csproj
```diff
@@ -48,6 +49,7 @@
   <ItemGroup>
     <PackageReference Include="DalamudPackager" Version="11.0.0" />
     <PackageReference Include="SharpDX.Mathematics" Version="4.2.0" />
+    <PackageReference Include="System.Data.Sqlite" Version="1.0.119" />
```
awgil, line 52: What code uses it?

BossMod/BossModule/ZoneModule.cs
awgil: I'd then make two independent functions (draw extra in main window & draw dedicated window).

BossMod/ActionQueue/ActionDefinition.cs
awgil: Ok, then why can't we just _always_ return `cdg.Total / MaxChargesAtCap() - cdg.Elapsed`, clamped to 0?

BossMod/BossModule/ZoneModule.cs
xanunderscore: Oh well yeah but we still need to know whether to display the hints window and the zone module window

BossMod/BossMod.csproj
xanunderscore: AutoClear.cs, scroll to the bottom

.github/workflows/build.yml
xanunderscore: removed, this was for my fork only, included accidentally

BossMod/ActionQueue/ActionDefinition.cs
xanunderscore: that should work in theory yeah. this code was originally only checking for specifically Drill, since it's the only action that is affected and i wanted to avoid introducing other bugs, but it seemed to work fine in practice

BossMod/Autorotation/MiscAI/AutoEngage.cs
xanunderscore: removed

PR #582, opened 2025-01-31
new base, DRK, & WAR implementation

Akechi-kun:
time to cook chief


BossMod/Autorotation/Standard/akechi/Tank/AkechiPLD.cs
```diff
@@ -2,7 +2,7 @@
 using AID = BossMod.PLD.AID;
 using SID = BossMod.PLD.SID;
 
-namespace BossMod.Autorotation.akechi;
+namespace BossMod.Autorotation.Standard.akechi.Tank;
```
awgil, line 5: Is this namespace change intended? It will break people's plans & presets.

PR #607, opened 2025-03-28
Underkeep Obstacle maps

erdelf:
the first two areas of underkeep

-3rd section could need it but is so vertical the obstacle maps I know how to generate are worthless


xanunderscore:
added a set of connected maps in 6e24153d38692f99b2436bab2c045e34328afb6a that should cover the entire third section properly

PR #621, opened 2025-04-23
Very basic TN and fallback fallback ranged option

Erisen8974:


BossMod/Autorotation/Standard/xan/Melee/VPR.cs
```diff
@@ -260,6 +260,8 @@ public override void Exec(StrategyValues strategy, Enemy? primaryTarget)
         // fallback for out of range
         if (Coil > 0)
             PushGCD(AID.UncoiledFury, BestRangedAOETarget);
+        else
+            PushGCD(AID.WrithingSnap, primaryTarget);
```
xanunderscore, line 264: add a strategy track for this. sometimes your GCD will hit 0 when you're, say, 3.2 units from the target, and it's better to wait that extra 0.1s to be in melee range than it is to roll an entire GCD using a ranged attack

BossMod/Autorotation/Standard/xan/Melee/VPR.cs
```diff
@@ -7,12 +7,25 @@ namespace BossMod.Autorotation.xan;
 
 public sealed class VPR(RotationModuleManager manager, Actor player) : Attackxan<AID, TraitID>(manager, player)
 {
+    public enum Track { WrithingSnap = SharedTrack.Count }
+
+    public enum WrithingSnapStrategy
+    {
+        Allow,
```
xanunderscore, line 14: make Forbid the default pls

xanunderscore:
Actually I'll save you the runaround and just implement these myself

xanunderscore:
Added in 98b2ac7bf8bf6fca7ace83cce95bacdb5f9bf071

PR #629, opened 2025-05-03
Module to stay out of the targets hitbox

Erisen8974:
This might also make more sense as another track in the stay close module instead.

This stays a distance away from the targets hitbox so we dont end up dodging to the dead center.

For bigger bosses like the arch-eta hunt being on the edge vs center of the hitbox can make the difference between being able to dodge the point blank in time or not.

### DIFF
--- a/BossMod/Autorotation/MiscAI/StayAwayFromTarget.cs
+++ b/BossMod/Autorotation/MiscAI/StayAwayFromTarget.cs
@@ -32,11 +32,11 @@ public sealed class StayAwayFromTarget(RotationModuleManager manager, Actor play
         if (primaryTarget != null)
         {
             var realDistance = strategy.Option(Tracks.Distance).Value.Option / 10f + primaryTarget.HitboxRadius;
-            var sqDist = 4 * realDistance * realDistance; // 4x to make .25 prio
+            var sqDist = 2 * realDistance * realDistance; // 2x to make .5 prio
             var tagetPosition = primaryTarget.Position;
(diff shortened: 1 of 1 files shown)


xanunderscore:
as discussed on discord, i don't think this is worth the additional maintenance overhead and debugging burden in exchange for making a single module for non-current content more reliable. i would rather add this hint to Arch-Eta specifically, or failing that, i would like more examples of bosses where "stay on edge of hitbox" is a good strategy (for example, it definitely isn't on the dawntrail tree guy A mark, and i feel like there's a squirrel or something that also spams directional cleaves)

Erisen8974:
I dont see any DT A rank that looks like a tree so I assume you mean the EW one Aegeiros. That hunt works fine staying on the edge and most of its cleaves come with a pointblank anyway. It also has player baited pointblanks that if baited dead center may force downtime (im not sure it might be small enough) but if on the edge has no issues.

There is also ovni which has concussive oscilation/mega storm that requires pretty quick in or out that is best solved from near hitbox.

There is a ShB A rank that has a similar player baited aoe to the tree except I know that one will force downtime if baited too close center.

Most mechs are solvable from the hitbox because thats where SE expects you to stand by default, frequently punishing being too close (big pb aoe) or too far (long cones).

But none of that really matters the main reason I want to be on the hitbox is that its more cozy and I dont like being in the middle

xanunderscore:
Treant was the term I was thinking of, not tree. Nechuciho was the guy. Eh whatever who cares. I'll add this to the stay close to target module.

PR #634, opened 2025-05-13
Autorotation window cleanup

Akechi-kun:
default
image
image

e.g. 1
image
image

e.g. 2
image
image


xanunderscore:
I'm gonna keep this open but i have my own pet ideas of making an ActionTimeline-esque timeline graphic to replace the queue window

xanunderscore:
Your logic here is actually not correct because `Entries` is unsorted, so what the window will display is whatever are the first GCD and oGCD that got added, regardless of which prio they are. (See `FindBest`.)

I'd like the autorot window to look more like the timeline view on e.g. xivanalysis, which is going to require a major overhaul, so I'm going to close this.

PR #641, opened 2025-05-31
Add a buffer hint to over dodge AOEs

Erisen8974:
Adds a configurable 0/.5/1.5/3 yalm buffer in the form of a small priority reduction near forbidden zones.

When set to none this should behave identically to without this change. 

In my testing I did not notice any frame loss from the extra pass. 30 fps with or without

### DIFF
--- a/BossMod/Autorotation/MiscAI/NormalMovement.cs
+++ b/BossMod/Autorotation/MiscAI/NormalMovement.cs
@@ -4,10 +4,11 @@ namespace BossMod.Autorotation.MiscAI;
 
 public sealed class NormalMovement(RotationModuleManager manager, Actor player) : RotationModule(manager, player)
 {
-    public enum Track { Destination, Range, Cast, SpecialModes }
+    public enum Track { Destination, Range, Cast, SpecialModes, ForbiddenZoneCushion }
     public enum DestinationStrategy { None, Pathfind, Explicit }
     public enum RangeStrategy { Any, MaxMelee, MeleeGreedGCDExplicit, MeleeGreedLastMomentExplicit }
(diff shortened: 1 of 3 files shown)


xanunderscore:
We had this functionality already in ShapeDistance https://github.com/awgil/ffxiv_bossmod/blob/ee5d5753b4add216d03a3308d88f2da61272e182/BossMod/Util/ShapeDistance.cs but I removed it since there wasn't any practical use for it at the time. Using signed distance fields has the benefit of expanding or contracting the forbidden zone without additional runtime cost, whereas it looks like this will result in a flat 9x multiplier to grid rasterization time. In either case, there are undesirable edge cases, like encounters with multiple nearby forbidden zones that are thin polygons (like FRU P1) where the zone cushion tends to swallow up the entire middle of the arena.

I'm open to reverting that commit but I would like a decent justification for doing so - are you just trying to make AI mode look less "suspicious"?

Erisen8974:
so I can bot in crescent without being too obvious:)

it shouldnt do 9x to the rasterization cost because it is not redoing the expensive operations, iterating all the shape functions, just a couple of array accesses with the precalculated values.  I tried with/without this and saw no frame rate diference.

The default is to do nothing identical to current behavior, this only expands the danger zones if actively set.

Since it is expanded via goal zone prio not forbidden zones if the over dodge is not possible it will just do the pixel dodge instead (no goal zones set so it just uses the forbidden zones). By expanding/contracting the shapes of the forbidden zone it would fail if the change results in no safe area.

xanunderscore:
Oh ok yeah it's not doing 9x I can't read

Erisen8974:
I come back with hard numbers for the overhead of the expansion. Assuming the visualizer updates I just pushed are measuring the right thing...

On my replay of the black chocobo fate it takes between 7 and 8 ms with the slider at 0 or with the slider above 0.

PR #644, opened 2025-06-06
Implicit greed option

Erisen8974:
An option to greed max melee until near when the attack goes off then usually probably get out in time

### DIFF
--- a/BossMod/Autorotation/MiscAI/NormalMovement.cs
+++ b/BossMod/Autorotation/MiscAI/NormalMovement.cs
@@ -135,10 +135,11 @@ public sealed class NormalMovement(RotationModuleManager manager, Actor player) 
                                 navi.Destination = uptimePosition;
                             break;
                         case RangeStrategy.MeleeGreedLastMomentImplicit:
-                            // Leeway is on the order of 1 second when in danger or near maxfloat when safe.
-                            // If leeway is over 10 minutes then the player is currently safe so dont greed INTO danger.
-                            // This avoids vibrating on the edge of the attack due to the maxfloat leeway being higher than 0.
-                            if (navi.LeewaySeconds is < 600 and > 0)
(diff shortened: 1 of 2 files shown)


xanunderscore:
what did you test this on?

BossMod/Autorotation/MiscAI/NormalMovement.cs
```diff
@@ -133,6 +134,21 @@ public override void Execute(StrategyValues strategy, ref Actor? primaryTarget,
                             if (navi.LeewaySeconds > (rangeStrategy == RangeStrategy.MeleeGreedGCDExplicit ? GCD : 0))
                                 navi.Destination = uptimePosition;
                             break;
+                        case RangeStrategy.MeleeGreedLastMomentImplicit:
+                            // If the uptime pixel G is lower than the current position G then dont greed INTO danger.
+                            // This avoids vibrating on the edge of the attack due to the maxfloat leeway of already being safe being higher than 0.
+                            var uptimeIndex = _navCtx.Map.GridToIndex(_navCtx.Map.WorldToGrid(uptimePosition));
+                            var currentIndex = _navCtx.ThetaStar._startNodeIndex;
+                            if (navi.LeewaySeconds > 0.5)
```
xanunderscore, line 142: How did you decide on 0.5s?

Erisen8974: by picking an arbitrary number higher than 0

The pathfinding seems to behave less well when it thinks there is no possible path (negative leeway) even when the inherent 1 second buffer would make it evade it would occasionally walk through an aoe that it "couldn't dodge". So I picked a small positive number so when its dodging time the leeway is still positive.

xanunderscore:
Adding this to my local build, will test today

PR #653, opened 2025-07-14
World hints for C033Statice Fireworks

oorzkws:
This PR adds world hints to the Fireworks mechanic on Another Aloalo Island/Statice.

I'm unsure if there's a nice way to cache the result considering it's the same between `DrawArenaForeground` and `AddMovementHints`; I'm open to feedback on this.


oorzkws:
Steps are now all laid out for chains:
- Stack (green)
- Break towards relative North (yellow)
- Move to safe spot (red)

Colors will progress so the current step is always green, next is yellow, etc.
image

There's also a relative north indicator for bombs, the triangle.

oorzkws:
Should be done adding things

xanunderscore:
PR looks good, I just want to redo this encounter for myself before i merge it to make sure the hints are intuitive enough

oorzkws:
Sounds good. If you want a duty recorder recording of the encounter I can send it your way privately as well (as it includes my IGN etc).

PR #654, opened 2025-07-16
IPC: Add `Configuration`, and `ActionQueue` methods

zbee:
This PR adds a few IPC methods to make it possible to check the `ActionQueue` entries, as well as when BossMod's configuration was last modified.

> The use-case here is for Wrath to be able to poll BossMod for queued actions, regardless of preset, and if actions are consistently going out, that itself can be marked as a preset (as opposed to the whole plugin).

- [X] Adds `Configuration.LastModified`, updated with the `Service.Config.Modified` `Event`.
- [X] Adds `Rotation.ActionQueue.HasEntries`, a `bool` indicating if there are any `Entry`s that are automatic actions.

### DIFF
--- a/BossMod/ActionQueue/ActionQueue.cs
+++ b/BossMod/ActionQueue/ActionQueue.cs
@@ -8,7 +8,10 @@
 // - repeat the process until no more actions can be found
 public sealed class ActionQueue
 {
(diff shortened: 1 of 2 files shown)


xanunderscore:
pls change the queue stuff so that it's just one IPC method that returns a number (or a bool) indicating whether there are non-manual entries in the queue, we don't need to go through the trouble of allocating a list for this

zbee:
@xanunderscore please let me know if this PR needs anything else.

PR #665, opened 2025-08-29
Qarn Update for 7.2 Changes

goatzonedotnet:
going to circle back to zone module later after more poking around.


Goatzone:
hang on: didn't set `arenacolor.enemy` for adds; need to test that, give a few.

PR #681, opened 2025-11-04
adding spawnroom flag to minimap

pepper-jelly:
So I checked texture and added new uv coordinates.


xanunderscore:
BTW I originally had chests at the "correct" scale, but I reduced it in order to be able to show up to 3 chest icons in one room. Ingame map only shows one chest and 3 chests don't really fit

PR #684, opened 2025-11-14
Add support for Great Hunt normal

alydevs:
Probably a bunch of unnecessary stuff in here, it's mostly copied from ex, but i changed the OID and AID enums and ran it a couple times, confirmed that it seems to work. unsync lv100 vpr


xanunderscore:
ported to master

PR #686, opened 2025-11-19
normalmove and therematch changes

Oblituarius:
Added logic to NormalMove. This will help to not spawn new pathfinding tasks if ForceMovement is set
Added new nav coords to TheRematch SoloPlayerDuty (QuestBattles) to improve navigation.


xanunderscore:
I made a similar change locally, it will be in the next update

PR #714, opened 2026-04-07
check for chocobo in stable before summoning

Jaksuhn:
if the chocobo is stabled, this locks the entire autorot/util


xanunderscore:
Note that when changing argument count of existing ops (but not when adding a new one) we need to make sure to preserve replay compatibility, otherwise we get data corruption when loading older replays. See https://github.com/awgil/ffxiv_bossmod/commit/ec4c93b885e36f15c5ca5fdce5430758d7baec27

Jaksuhn:
Noted. Thank you!

PR #716, opened 2026-04-10
tender valley support

cerq-dev:
Supports the 3 bosses, additionaly for the serpent, when it casts greatest labyrinth, it shows the tile you should walk to.

The only thing left to do is add support for ancient wrath from the mini-boss.

There are a few ai hints missing but since this dungeon does not have duty support/trusts I think it'll be just people running it once for completition and not  with AD (i could try adding them if its worth it/needed)


xanunderscore:
It was marked as draft, do you want it merged?

PR #722, opened 2026-04-28
Add Module for Normal Everkeep

gadeleon:
(This is my first attempt at a module and should definitely be squashed if pulled.)

Adds support for Normal Everkeep for Duty Support.

Tested as:
* GNB
* WHM
* BLM
* BRD
* SAM

Zero hits on all runs.

Open to feedback to make this better!

### DIFF
--- a/BossMod/Modules/Dawntrail/Trial/T03Everkeep/BitterReaping.cs
+++ b/BossMod/Modules/Dawntrail/Trial/T03Everkeep/BitterReaping.cs
@@ -1,0 +1,3 @@
+namespace BossMod.Dawntrail.Trial.T03Everkeep;
+
+class BitterReaping(BossModule module) : Components.SingleTargetCast(module, AID.BitterReapingAOE, "Tank busters");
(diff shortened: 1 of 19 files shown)


xanunderscore:
Will try to get this merged after I update the plugin for 7.5, which is blocked on a staging release of Dalamud. Thanks for the PR

xanunderscore:
Ok finally getting to it now lel

BossMod/Modules/Dawntrail/Trial/T03Everkeep/BitterReaping.cs
```diff
@@ -0,0 +1,3 @@
+namespace BossMod.Dawntrail.Trial.T03Everkeep;
+
+class BitterReaping(BossModule module) : Components.SingleTargetCast(module, AID.BitterReapingAOE, "Tank busters");
```
xanunderscore, line 3: For components that are a single line, feel free to just put them in the root module file (T03Everkeep), creating separate files for simple components is more trouble than it's worth I feel (I will merge this as-is though)

BossMod/Modules/Dawntrail/Trial/T03Everkeep/FireIII.cs
```diff
@@ -0,0 +1,15 @@
+namespace BossMod.Dawntrail.Trial.T03Everkeep;
+
+// Spread on icon 376; 8 helpers instant-cast Fire III (37752) ~5s later. Forbidden radius is 7m
+// (2m wider than the 5m hit) so the AI doesn't clip a neighbor while routing.
+class FireIII(BossModule module) : Components.SpreadFromIcon(module, 376, AID.FireIII, 7f, 5.1f)
```
xanunderscore, line 5: GenericStackSpread already has an extra spread threshold for AI (in this case 1 unit) so the extra size here should be unnecessary

xanunderscore (approved):

xanunderscore:
Merged in 275db9712d879bc0960223cf1b200f2575a45a6a

PR #730, opened 2026-05-10
Fix: CFC ID/Enforce Snap Punch for first phase of Return of the Holyfist.

gadeleon:
I was having trouble with this QuestBattle and noticed the CFC was off by 1.

```
15:11:22.262 | DBG | [ClientState] ZoneInit: ZoneInitEventArgs { TerritoryTypeId = 261, Instance = 0, ContentFinderCondition = 322, Weather = 14, ActiveFestivals = [0|0, 0|0, 0|0, 0|0, 0|0, 0|0, 0|0, 0|0],  }
```
On top of that, I was overlevelled/overgeared and kept reducing Holyfist's health to 1 which made the auto rotation stop/no actions getting cast so Snap Punch never executed 3 times.

This forces the use of the combo (without demolition) to complete the first phase.

### DIFF
--- a/BossMod/QuestBattle/ARealmReborn/ClassJobQuests/MNK/ReturnoftheHolyfist.cs
+++ b/BossMod/QuestBattle/ARealmReborn/ClassJobQuests/MNK/ReturnoftheHolyfist.cs
@@ -1,7 +1,47 @@
-﻿namespace BossMod.QuestBattle.ARealmReborn.ClassJobQuests.MNK;
+namespace BossMod.QuestBattle.ARealmReborn.ClassJobQuests.MNK;
 
-[ZoneModuleInfo(BossModuleInfo.Maturity.Contributed, 323)]
+[ZoneModuleInfo(BossModuleInfo.Maturity.Contributed, 322, 261)]
 internal class ReturnOfTheHolyfist(WorldState ws) : QuestBattle(ws)
 {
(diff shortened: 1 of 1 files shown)


BossMod/QuestBattle/ARealmReborn/ClassJobQuests/MNK/ReturnoftheHolyfist.cs
```diff
@@ -1,7 +1,47 @@
-﻿namespace BossMod.QuestBattle.ARealmReborn.ClassJobQuests.MNK;
+namespace BossMod.QuestBattle.ARealmReborn.ClassJobQuests.MNK;
 
-[ZoneModuleInfo(BossModuleInfo.Maturity.Contributed, 323)]
+[ZoneModuleInfo(BossModuleInfo.Maturity.Contributed, 322, 261)]
```
xanunderscore, line 3: Not sure if this is correct. 322 and 323 both map to quest battles that are called "Return of the Holyfist". I don't know why there are two of them.

xanunderscore:
This seems like overkill to me. If you force his priority to 0 then the MNK rotation will naturally use snap punch

gadeleon:
> This seems like overkill to me. If you force his priority to 0 then the MNK rotation will naturally use snap punch

Will give it a shot. My issue was that when his health reached 1, BossMod stopped doing actions.

PR #731, opened 2026-05-13
Feat: Add Our Compromise Quest Battle

gadeleon:
All Enemy NPCs in this Quest Battle are L??. It seems BossMod's default behavior is to ignore any L?? mobs. This module creates AI Hints to force priority of NPCs which kicks the AI to attack them.


xanunderscore:
Will investigate. `L??` is UI-only and has nothing to do with prioritization. (You can check in vbm's object debugger to see what their actual level is.)

PR #742, opened 2026-07-06
AkechiPLD improvements

### LINKED ISSUE #741: [BUG] L100 Akechi PLD: : Requiescat planner timing/range behavior
1. In the Req track, some options use a 180s cooldown display even though its 60s

From what I can tell:
Together is listed as 60s
RaidBuffsOnly, Force, and ForceWeave are 180s

2. Imperator should be usable outside melee range

Imperator can be used at range, but currently the logic requires the target to be within 3y. This can cause Imperator not to be used and drift.  I think the issue may be that Requiescat and Imperator share the same planner track, but the module’s usage condition treats both as if they require melee range. 
A possible fix would be to keep the same Req. track, but base the range check on BestRequiescat: use 3y when BestRequiescat is AID.Requiescat, and 25y when it is AID.Imperator. 

3. Blades and Holy Spirit

The blade combo can be dropped if Requiescat is no longer active, even when there are steps available. 
Happens if req falls due to small downtime, but blades are still available)

Expected:
Req / Imperator required to start Confiteor.
Once the Blade combo has started, the module should continue Blade of Faith / Truth / Valor based on BladeComboStep

Holy Spirit
Currently, Divine Might Holy Spirit is spent immediately after the Atonement chain. Ideally yo want to spend Holy Spirit right before the next Royal Authority would overwrite Divine Might.

Desired behavior:
Royal Authority → Atonement → Supplication → Sepulchre → Fast Blade → Riot Blade → Holy Spirit → Royal Authority

It should still be used early if:
the player is outside melee range,
Divine Might is about to expire,
Divine Might would be overwritten by the next Royal Authority,
or the user explicitly forces Holy Spirit.

But normally DMHS should probably be held until the next combo GCD would be Royal Authority.

Ik its lowk alot- I can try to fork and submit a PR with these changes if that would be helpful.

Akechi-kun:
Closes #741 

- [x] Fixed issue with `Requiescat` cooldown value being incorrect in cooldown planner
- [x] Fixed issue with `Imperator` only being used while in melee range
- [x] Fixed issue with `Imperator` targeting 
- Now correctly aims for best splash target
- [x] Added `VeryLate` option to `HolyStrategy` 
- Can now use right before next Divine Might grant
- [x] Refactored options for `DashStrategy`
- `Automatic`, `>3y`,`>5y`, `>10y`, Opener only, Opener only `>3y`, overcap safely, overcap aggressively
- [x] Removed `Early` option from `GoringBladeStrategy`
- Unnecessary as `Automatic` behavior is literally the same thing
- [x] Definitions cleanup
- [x] Spaghetti cleanup

### DIFF
--- a/BossMod/Autorotation/Standard/akechi/AkechiTools.cs
+++ b/BossMod/Autorotation/Standard/akechi/AkechiTools.cs
@@ -128,6 +128,7 @@ public abstract class AkechiTools<AID, TraitID>(RotationModuleManager manager, A
     protected bool Unlocked(AID aid) => ActionUnlocked(ActionID.MakeSpell(aid));
     protected bool Unlocked(TraitID tid) => TraitUnlocked((uint)(object)tid);
     protected AID ComboLastMove => (AID)(object)World.Client.ComboState.Action;
+    protected unsafe AID LastComboAction => (AID)(object)Instance()->Combo.Action;
     protected float SkSGCDLength => ActionSpeed.GCDRounded(World.Client.PlayerStats.SkillSpeed, World.Client.PlayerStats.Haste, Player.Level);
     protected float SpSGCDLength => ActionSpeed.GCDRounded(World.Client.PlayerStats.SpellSpeed, World.Client.PlayerStats.Haste, Player.Level);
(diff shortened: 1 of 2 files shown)


prompt-to-code:
You're amazing, thank you! Will test soon

xanunderscore:
Merged into testing

PR #744, opened 2026-07-08
The Merchant's Tale (Advanced) - DaryaTheSeaMaid 

Equilius:
Hey, started working on DaryaTheSeaMaid (advanced) since I've seen people ask about it every so often. Once it is ready, I will mark it as ready for review.


Equilius:
The module has been completed; I tested it in a 4-man party with AI on, and it worked fine. Just wanna test it in a 2-man party since some mechanics get changed around, and if that works fine, then I will change it from draft to ready for review.

Equilius:
Going to stop adding to this now; I just wanted to learn about using AI for a module, and I think I went a bit overkill with it since no one is going to use AI for this anyway, since it is a matchmade instance

PR #745, opened 2026-07-12
OC South horn fates

Equilius:
Someone was asking about OC South Horn's normal fates, so I decided to add them.

I have tested each fate at least 2x, and everything seems to be working; However, some spell IDs might be missing, as it's impossible to make the fate last long enough to get the spell ID and see what it does, since other people will just nuke the boss.


Equilius:
Thanks for the help in figuring out what spells are missing from the Fates. I will continue working on this now and hopefully have it ready soon!

PR #746, opened 2026-07-22
AnotherMountRokkon cleanup

VeraNala:
I may or may not be running this content atm,
New on Left, Old on Right
ffxiv_dx11_a4ERHvLDem


xanunderscore:
Merged into testing

PR #747, opened 2026-07-26
add bulk replay cull

Jaksuhn:
little ui additions in the replay window to allow you to delete replays en masse by either duty or duration


xanunderscore:
Merged into testing

PR #755, opened 2026-08-06
North horn normal fates

Equilius:
All normal fates which have been tested multiple times with AI on


xanunderscore:
Merged into testing, ty

PR #758, opened 2026-08-11
Adds edge cushion for more positional margin

pedro-bale:
Same context as overdodge for Automatic Movement, but here we can go more towards the center of the positional to have more margin. This is very useful specially with ping, since we can account a few more frames to be in the correct positional before the boss moves, and it's less likely to fail the positional since without this, a micro movement from boss could lead to not being in the correct positional edge anymore.

### DIFF
--- a/BossMod/Autorotation/MiscAI/GoToPositional.cs
+++ b/BossMod/Autorotation/MiscAI/GoToPositional.cs
@@ -4,8 +4,11 @@ public sealed class GoToPositional(RotationModuleManager manager, Actor player) 
 {
     public enum Tracks
     {
-        Positional
+        Positional,
+        EdgeBuffer
     }
+
+    public enum EdgeBufferStrategy { None, Small, Medium, Large }
 
     public static RotationModuleDefinition Definition()
     {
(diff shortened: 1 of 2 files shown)


BossMod/Autorotation/MiscAI/GoToPositional.cs
```diff
@@ -17,6 +20,13 @@ public static RotationModuleDefinition Definition()
         {
             track.AddOption(positional);
         }
+
+        def.Define(Tracks.EdgeBuffer).As<EdgeBufferStrategy>("EdgeBuffer", "Edge buffer", 20)
+            .AddOption(EdgeBufferStrategy.None, "Stand at positional edges")
+            .AddOption(EdgeBufferStrategy.Small, "Prefer staying 0.5y inside from the edges")
```
pedro-bale, line 26: I'm not sure 0.5y is a relevant distance to keep here, but 🤷‍♂️

xanunderscore:
Not really a fan of this implementation. I would prefer to do the inverse of the forbidden zone cushion logic in NavigationDecision, which would apply to all goalzones, not just positionals (i.e. encouraging melees to stay slightly closer to the boss).

xanunderscore:
Actually this should be fine

PR #768, opened 2026-09-13
First Crucible Board

VeraNala:
BST/Crucible wiring and First Board.
Idk if this naming schema works for you but lmk


xanunderscore:
Plz rename the modules to include the boss name in them, I think that's easier to read. Also just `Crucible` is fine for the category.

